PR #6, opened 2023-05-11
Added error codes and improved details by including part of the value…

abergs:
… that was sent

### DIFF
--- a/src/Service/TokenService.cs
+++ b/src/Service/TokenService.cs
@@ -82,7 +82,7 @@ public class TokenService
 
     public T DecodeToken<T>(string token, string prefix, bool contractless = false)
     {
-        if (token == null) throw new ArgumentNullException("token");
+        if (token == null) throw new ApiException("missing_token", $"This operation requires a token that starts with '{prefix}' to be passed.",400);
 
         if (prefix != null)
         {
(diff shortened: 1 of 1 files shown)


src/Service/TokenService.cs
```diff
@@ -92,9 +92,9 @@ public T DecodeToken<T>(string token, string prefix, bool contractless = false)
             }
             else
             {
-                _log.LogWarning("Could not remove prefix={prefix}", prefix);
-                // todo: Uncomment to require new token format. Should be safe to do after 2020-01-30:
-                throw new ApiException($"The token you sent was not correct. The token used for this endpoint should start with '{prefix}'. Make sure you are not sending the wrong value.", 400);
+                var invalidInput = token[..Math.Min(10, token.Length)];
+                _log.LogWarning("Could not remove prefix={prefix}, token started with {InvalidInput}", prefix, invalidInput);
```
justindbaur, line 96: nit: That we are mixing camelCase and PascalCase for the named holes but I know we this isn't the only place and part of why I might turn on this rule soon: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1727

justindbaur (approved):

src/Service/TokenService.cs
abergs: I agree that we should turn on that rule. @justindbaur

PR #16, opened 2023-05-15
Added test and format workflow

abergs:


.github/workflows/test-and-format.yml
```diff
@@ -0,0 +1,50 @@
...
+      - uses: actions/checkout@v3
+      - name: Setup dotnet
+        uses: actions/setup-dotnet@v3
+        # with:
+        #   dotnet-version: ${{ matrix.dotnet-version }}
+      # You can test your matrix by printing the current dotnet version
+      - name: Display dotnet version
+        run: dotnet --version
+      - name: Install dependencies
+        run: dotnet restore --locked-mode
```
justindbaur, line 32: This repo doesn't use lock files, or is this in preparation to do it in the future?

abergs: Since the build works regardless I thought i could just as well use it?

.github/workflows/test-and-format.yml
```diff
@@ -0,0 +1,50 @@
...
+
+    runs-on: ubuntu-latest
+    # strategy:
+    #   matrix:
+    #     dotnet-version: [ '3.1.x', '6.0.x' ]
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Setup dotnet
+        uses: actions/setup-dotnet@v3
```
joseph-flinn, line 25: We have adopted the pattern of pinning the Action hash instead of the major/minor version to mitigate the risk of supply chain attacks in the pipeline dumping our repo secrets.

abergs: Added

joseph-flinn (approved):
Looks good to me!

.github/workflows/test-and-format.yml
justindbaur: Yeah that's fine. Just wondering if you also meant to introduce them.

justindbaur (approved):

PR #17, opened 2023-05-16
Improve Unknown Credential Exception

abergs:
- Introduce Common UnknownCredentialException to improve reporting
- Add tests

For reporting reasons it would be helpful to separate this exception into it's own type.
I've also added tests to confirm it's being caught and handled correctly.

### DIFF
--- a/src/Service/Helpers/UnknownCredentialException.cs
+++ b/src/Service/Helpers/UnknownCredentialException.cs
@@ -1,6 +1,7 @@
 namespace Passwordless.Service.Helpers;
 
-public class UnknownCredentialException : ApiException {
+public class UnknownCredentialException : ApiException
+{
     public UnknownCredentialException(string message) : base(message)
     {
     }
--- a/src/Service/Helpers/UnknownCredentialException.cs
+++ b/src/Service/Helpers/UnknownCredentialException.cs
@@ -3,7 +3,7 @@ namespace Passwordless.Service.Helpers;
 public class UnknownCredentialException : ApiException
 {
     public UnknownCredentialException(string credentialId)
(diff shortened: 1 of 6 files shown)


src/Service/Helpers/UnknownCredentialException.cs
```diff
@@ -0,0 +1,16 @@
+namespace Passwordless.Service.Helpers;
+
+public class UnknownCredentialException : ApiException
+{
+    public UnknownCredentialException(string message) : base(message)
+    {
+    }
+
+    public UnknownCredentialException(string message, int statusCode, Dictionary<string, object> extras = null) : base(message, statusCode, extras)
```
justindbaur, line 9: Personally for each exception that wraps `ApiException` I would have it not simply expose the base class constructors and instead be like:
```C#
public UnknownCredentialException(string credentialId)
: base("unknown_credential", "We don't recognize the passkey you sent us", 400, new () { { "credentialId", credentialId }})
{
    CredentialId = credentialId;
}
```
Maybe it could have another constructor that allows customizing the message but for the most part it is a helper around `ApiException`. Let me know if you disagree though.

tests/Api.IntegrationTests/BasicTests.cs
```diff
@@ -13,11 +13,27 @@ public BasicTests(TestWebApplicationFactory<Program> factory)
...
+        _client.DefaultRequestHeaders.Add("ApiSecret", _factory.ApiSecret);
+    }
+}
+
+public class PublicTests : BasicTests
+{
+    public PublicTests(TestWebApplicationFactory<Program> factory) : base(factory)
+    {
+        _client.DefaultRequestHeaders.Add("ApiKey", _factory.ApiKey);
+    }
```
justindbaur, line 38: Nice!

PR #18, opened 2023-05-16
Move Alias creation to register/complete

abergs:
1. register/token - verify if aliases is available when creating the token
2. register/token -  Change token type to include aliases
3. register/begin - send the aliases in the session
4. reigser/complete - retrieve them and store them on register complete.


justindbaur (approved):
Ship it!

PR #23, opened 2023-05-17
Make Fido2 Service More Testable

justindbaur:
- Make Fido2ServiceEndpoints a bit more testable by making a `IFido2ServiceFactory` that can create the service but also be mocked in tests to return a mocked version of `IFido2Service`. 
- Made an interface for `ITokenService` so that it can be mocked inside of `IFido2Service`
- Start a `Service` test project


src/Service/IFido2ServiceFactory.cs
```diff
@@ -0,0 +1,41 @@
...
+using Passwordless.Service.Storage;
+
+namespace Passwordless.Service;
+
+public interface IFido2ServiceFactory
+{
+    Task<IFido2Service> CreateAsync();
+}
+
+public class DefaultFido2ServiceFactory : IFido2ServiceFactory
```
abergs, line 12: Looks good. But if we would refactor the TokenService to not do a Init, we could get rid of this Factory, yes?

src/Service/IFido2Service.cs
```diff
@@ -0,0 +1,17 @@
+using Fido2NetLib;
+using Passwordless.Service.Models;
+using Passwordless.Service.Storage.Ef;
+
+namespace Passwordless.Service;
+
+public interface IFido2Service
```
abergs, line 7: Nice

src/Api/Endpoints/Signin.cs
```diff
@@ -13,39 +11,33 @@ public static class SigninEndpoints
...
+            var result = await fido2Service.SignInComplete(payload, deviceInfo, country);
 
             return Ok(result);
         })
             .RequirePublicKey()
             .RequireCors("default")
             .WithMetadata(new HttpMethodMetadata(new string[] { "POST" }, acceptCorsPreflight: true)); ;
 
-        app.MapPost("/signin/verify", async (SignInVerifyDTO payload, ClaimsPrincipal user, ITenantStorage storage) =>
+        app.MapPost("/signin/verify", async (SignInVerifyDTO payload, IFido2ServiceFactory fido2ServiceFactory) =>
```
abergs, line 37: Nice

abergs (approved):
Looks good

abergs (approved):
Looks good

src/Service/IFido2ServiceFactory.cs
justindbaur: Yep, and then we just replace injecting the factory and directly inject `IFido2Service`. It's the nicer solution but causes a lot more behavioral changes than I was comfortable doing.

PR #30, opened 2023-05-18
even easier development

abergs:
- Make Developing a bit Easier
- Formatting
- Fixed https and port
- Added better mail UX
- Added experimental support for doing migrations in the API project

### DIFF
--- a/src/Api/Helpers/AddDatabaseExtensionMethod.cs
+++ b/src/Api/Helpers/AddDatabaseExtensionMethod.cs
@@ -11,7 +11,7 @@ namespace Passwordless.Api.Helpers;
 
 public static class AddDatabaseExtensionMethod
 {
-    public static IServiceCollection AddDatabase(this IServiceCollection services, IConfiguration configuration)
+    public static IServiceCollection AddDatabase(this IServiceCollection services, IConfiguration configuration, IWebHostEnvironment environment)
     {
         // Database information
         var sqlite = configuration.GetConnectionString("sqlite");
@@ -41,8 +41,14 @@ public static class AddDatabaseExtensionMethod
 
         services.AddScoped<ITenantProvider>(sp =>
         {
-            var accountName = sp.GetService<IHttpContextAccessor>()?.HttpContext?.User
+            var context = sp.GetService<IHttpContextAccessor>()?.HttpContext;
+            var accountName = context?.User
                 .FindFirstValue(CustomClaimTypes.AccountName);
+            
+            
+            if(environment.IsDevelopment() && (context?.Request.Path == "/health/storage"|| context?.Request.Path == "/ApplyDatabaseMigrations")) {
+                return new ManualTenantProvider(accountName);
+            }
 
             return !string.IsNullOrEmpty(accountName)
                 ? new ManualTenantProvider(accountName)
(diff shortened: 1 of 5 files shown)


src/Api/Helpers/FriendlyExceptionsMiddleware.cs
```diff
@@ -21,10 +22,17 @@ public Task InvokeAsync(HttpContext context)
...
+
         try
         {
             await _next(context);
         }
+        catch (SqliteException e) when (env.IsDevelopment())
+        {
+            // this is done to allow migrations to be applied in development mode
+            throw;
+        }
```
justindbaur, line 35: Very nice, I tried to make the migrations endpoint in Api to but this was the solution I didn't think of, it was always being grabbed by our safe one. This doesn't work for when you use MS SQL for development but we could add a similar exception but I think that's fine for now. Sqlite will be the default dev DB so this is great.

src/Api/Helpers/AddDatabaseExtensionMethod.cs
```diff
@@ -41,9 +41,16 @@ public static IServiceCollection AddDatabase(this IServiceCollection services, I
 
         services.AddScoped<ITenantProvider>(sp =>
         {
-            var accountName = sp.GetService<IHttpContextAccessor>()?.HttpContext?.User
+            var context = sp.GetService<IHttpContextAccessor>()?.HttpContext;
+            var accountName = context?.User
                 .FindFirstValue(CustomClaimTypes.AccountName);
 
+
+            if (environment.IsDevelopment() && (context?.Request.Path == "/health/storage" || context?.Request.Path == "/ApplyDatabaseMigrations"))
```
justindbaur, line 49: What I was thinking is we could have a environment conditional `MapGet("/")` so that when in development it injects `ITenantStorageFactory` and we just attempt to do something minor in the database. This was the application would launch to it and if there are database issues it will show the fancy exception page. 

But in non-dev it does what it does now.

justindbaur: In that case we wouldn't need a seperate check for `Path == "/ApplyDatabaseMigrations"`

src/Api/Helpers/AddDatabaseExtensionMethod.cs
```diff
@@ -41,9 +41,16 @@ public static IServiceCollection AddDatabase(this IServiceCollection services, I
 
         services.AddScoped<ITenantProvider>(sp =>
         {
-            var accountName = sp.GetService<IHttpContextAccessor>()?.HttpContext?.User
+            var context = sp.GetService<IHttpContextAccessor>()?.HttpContext;
+            var accountName = context?.User
                 .FindFirstValue(CustomClaimTypes.AccountName);
 
+
```
justindbaur, line 48: Also, if we need to do an environment check I would rather get it from `sp.GetRequiredService<IWebHostEnvironment>()` so we can avoid outer scope capture.

src/Api/Helpers/AddDatabaseExtensionMethod.cs
abergs: @justindbaur Smart idea with the `/` path, but we would probably still need to whitelist `/ApplyDatabaseMigrations` because the button you click in the DatabaseExceptionHandler calls the `/ApplyDatabaseMigrations` -- are you implying we call `context.MigrateAsync` ourself?

justindbaur: Resolved in a call, I wasn't aware `/ApplyDatabaseMigrations` was a post under the hood by the migrations exception page. We did decide to run it ourselves in root because we also want to seed the database for easier development anyways.

justindbaur (approved):

PR #34, opened 2023-05-19
Test ci

joseph-flinn:
- Run Tests on every pushed commit
- Run Tests and Builds on every push to `master` and every `pull-request`


joseph-flinn:
Not ready yet

PR #35, opened 2023-05-19
Add Friendlier Exception From SDK

justindbaur:
- Adds an inner `DelegateHandler` so that all non-successful responses can be parsed into problem details and thrown as an exception. This means that it does throw more often than it did before.


abergs:
@justindbaur Looks great! I had a local branch that I didn't have time to finish - and overall your details are better.

abergs (approved):

abergs (approved):

PR #38, opened 2023-06-01
Changes to `Passwordless.AspNetCore`

justindbaur:
- Switch to using Endpoints so that our code behaves more like other user code
  - Also gives the consumer the ability to add CORS protection onto endpoints
- Remove need for full `ASP.NET Identity`.
  - Now only `UserStore<>` and `IUserClaimsPrincipalFactory<>` are needed (included in Identity).
- Allow customization of request options for all requests and per request.
- Add a register endpoint (off by default)
  - Has some customizability, more can be added after use cases have been seen
- Added XML documentation
- Added debug helpers
- Log for easier debugging


src/AdminConsole/Pages/Account/Login.cshtml
```diff
@@ -82,11 +82,12 @@
             setup() {
 
                 async function signIn(token) {
-                    var res = await fetch("/passwordless/verify-signin", {
+                    var res = await fetch("/passwordless-login", {
                         method: "POST",
-                        body: new URLSearchParams({ token }),
```
abergs, line 87: While not critical, I find using URLSearchParams to be a bit easier than JSON.

It doesn't require changing ContentType and you can also bind to non-complex objects on the endpoints, such as (string token) without the wrapper DTO.

abergs (commented):
I need to clone and run to do a full review. But on first review, it looks good. Anything you want me to focus on?

justindbaur:
@abergs Especially the defaults defined in `src/AspNetCore/PasswordlessRegisterOptions.cs` and just the general developer experience you think you'd get from this.

abergs:
@justindbaur I'm very close to just merging this and then doing discussions in another PR, but I was pondering if our naming can cause confusion.

In Passwordless docs and JS client we've reserved "Register" to mean "Register a credential to a user" and in the asp.net framework we're slightly adjusting it to mean "Register a credential to a new user" while existing users are served by "add-credential".

Would it make more sense structure endpoints to be:
* LoginUser
* RegisterNewUser
* RegisterCredential

If we don't change, the terminology gets mixed and easily confused in the docs:

"Get a register_token (see Register Options) and call the AddNewCredential-enpoint (Not the register endpoint).

PR #39, opened 2023-06-01
Change Handler Lifetime

justindbaur:
This is a fun one, I can't actually reproduce the error from using the Sdk in AdminConsole or in a test environment so maybe this isn't a bug. 

When I consume `Passwordless.AspNetCore` (which consumes `Passwordless.Net`) in a personal project I get this error: `The 'InnerHandler' property must be null. 'DelegatingHandler' instances provided to 'HttpMessageHandlerBuilder' must not be reused or cached`. 

I was pretty sure I didn't have to worry about the scope of my handler because it is stateless and has no dependencies like mentioned here: https://andrewlock.net/understanding-scopes-with-ihttpclientfactory-message-handlers/#does-having-mis-matched-scopes-matter- but I was having issues anyways. If I switched to using `AddTransient` the bug stopped happening in my project. 

Also, any instances [I find](https://grep.app/search?q=.AddHttpMessageHandler%3C) of this call on grep.app show registering the service as transient so I think it's a good idea.


abergs:
Alright. Here's some more confirmation to support your findings: https://stackoverflow.com/a/68867517

justindbaur:
@abergs Interesting, that answer links to Andrew Lock's article but it actually made me think the scope _shouldn't_ matter. Glad that answer agrees with my findings though. Still don't understand why I couldn't replicate the error in a test though.

PR #40, opened 2023-06-09
PAS173: Self-Hosting - Bootstrapping API & AdminConsole

kspearrin:


abergs:
Startup now works -- however, we still run into a bit of bootstrapping problem when booting the AdminConsole because we want to connect to the API, and we don't have valid API keys.

src/AdminConsole/ApiHelper.cs
```diff
@@ -24,6 +24,16 @@
~
~public class PasswordlessManagementOptions
 {
     public string ApiUrl { get; set; }
     public string ManagementKey { get; set; }
+
+    private string? _publicApiUrl;
+    public string PublicApiUrl
```
abergs, line 31: I don't think this is needed. We should probably use the "ApiUrl" of the PasswordlessOptions instead, as this configuration is just for the endpoints where ManagementKey is used

docker/docker-compose.yml
```diff
@@ -7,48 +7,24 @@ services:
       - db
     env_file:
       - settings.env
-    image: ${REGISTRY:-bitwarden}/passwordless:${TAG:-beta}
+    image: ${REGISTRY:-bitwarden}/passwordless:${TAG:-latest}
     restart: always
     ports:
-      - "5000:5000"
-      - "5001:5001"
+      - "8443:8443"
```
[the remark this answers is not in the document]
jonashendrickx: Moved it to 5701

abergs:
@jonashendrickx I noticed two conflicts here, just pinging you for visibility.

docker/entrypoint.sh
```diff
@@ -0,0 +1,193 @@
...
+  scheme="http"
+fi
+if [ "$BWP_PORT" == "null" ]; then
+  echo "WARNING: 'BWP_PORT' not set, defaulting to 5701.";
+  exit 1; 
+fi
+
+export PasswordlessManagement__PublicApiUrl="$scheme://${BWP_DOMAIN:-localhost}:${BWP_PORT:-5701}/api";
+echo "[Configuration] API private: $PasswordlessManagement__ApiUrl";
+echo "[Configuration] API public: $PasswordlessManagement__PublicApiUrl";
```
justindbaur, line 107: I'm curious on what the point in having the, what I think is only the internal PORT, customizable. This doesn't change the port exposed by the container, that it done in the docker run command ex. `-p 1200:5701`. 

Probably more of a question for @kspearrin as I see this is also possible in Bitwarden unified.

jonashendrickx:
> @jonashendrickx I noticed two conflicts here, just pinging you for visibility.

Fixed!

PR #41, opened 2023-06-13
add mailkit smtp provider

kspearrin:
PR adds support for sending mail with SMTP configuration using mailkit library.


kspearrin:
I dont know why this is failing. I run `dotnet format` locally and there are no problems reported.

abergs (approved):

PR #42, opened 2023-06-13
DB support for self hosting

abergs:
- Added support for non conflicting connection strings
- Added support for automatic migration

### DIFF
--- a/src/Api/Helpers/AddDatabaseExtensionMethod.cs
+++ b/src/Api/Helpers/AddDatabaseExtensionMethod.cs
@@ -16,6 +16,10 @@ public static class AddDatabaseExtensionMethod
         // Database information
         var sqlite = configuration.GetConnectionString("sqlite");
         var mssql = configuration.GetConnectionString("mssql");
+        
+        sqlite = configuration.GetValue<string>("ConnectionStrings.Sqlite.Api", sqlite);
+        mssql = configuration.GetValue<string>("ConnectionStrings.Mssql.Api", mssql);
+        
         if (!string.IsNullOrEmpty(sqlite))
         {
             services.AddDbContext<DbTenantContext, SqliteContext>((sp, builder) =>
--- a/src/Api/Program.cs
+++ b/src/Api/Program.cs
@@ -1,11 +1,13 @@
 using System.Reflection;
 using System.Text.Json;
+using Microsoft.EntityFrameworkCore;
 using Passwordless.Api;
 using Passwordless.Api.Authorization;
 using Passwordless.Api.Endpoints;
 using Passwordless.Api.Helpers;
 using Passwordless.Server.Endpoints;
 using Passwordless.Service;
+using Passwordless.Service.Storage.Ef;
 using Serilog;
 using Serilog.Sinks.Datadog.Logs;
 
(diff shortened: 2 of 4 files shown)


src/Api/Program.cs
```diff
@@ -99,6 +101,15 @@
             "Hey, this place is for computers. Check out our human documentation instead: https://docs.passwordless.dev");
 }
 
+if (builder.Configuration.GetValue<bool>("SelfHosted"))
+{
+    // Migrate latest database changes during startup
+    using var scope = app.Services.CreateScope();
+    var dbContext = scope.ServiceProvider
+        .GetRequiredService<DbTenantContext>();
+    dbContext.Database.Migrate();
```
abergs, line 110: For this to work, comment 1 must be resolved.

src/Api/Helpers/AddDatabaseExtensionMethod.cs
```diff
@@ -47,6 +51,7 @@ public static IServiceCollection AddDatabase(this IServiceCollection services, I
 
             var environment = sp.GetRequiredService<IWebHostEnvironment>();
 
+            // need to add exception here to allow self hosted migrations or another way to initiate context
```
abergs, line 54: comment 1: Currently, we cannot resolve the ITenantProvider when getting the DBTenantContext during SelfHosted Startup.

Not sure how to solve this in a clean way. Currently I feel like manually initiating it is the cleanest way, which is telling. @justindbaur Any ideas?

Or should we change the way we're applying migrations for self hosted production startup? We want to make it automatic.

justindbaur: I have no idea what the best route forward is, so I laid out a few options that came to mind. Some easier, some that might be better architected. For sure not an exhaustive list though.

Option 1: Manually create the `DbTenantContext` by pushing in a fake `ITenantProvider`, the issue is this might not be as simple as you might like, I think you will be required to create the provider specific version. Otherwise migrations won't work, check on that though, I might be wrong. You'll have to do it like this basically. https://github.com/passwordless/passwordless-server/blob/1b9f4c17654739c5c7d00097e84456e19e021001/src/Service/Storage/EF/DbTenantFactory.cs#L22

Option 2: Install the `dotnet-ef` tool in the docker file and run `dotnet ef database update` based on user configuration in `entrypoint.sh`. I'm not actually sure if this is possible, because you'd be doing this on a `Passwordless.Api.dll` and not on the csproj. That command has a `--no-build` command but I think this is still stretching possibilities. Worth investigating further though.

Option 3: Do it in a IHostedService with it's own service provider scope, then run migrations. (This could be done in startup too)

Option 4: Totally move away from the `DbTenantContext` being responsible for the tenant separation and instead it's "optional" and enforced on a per query basis. 

Option 5: Have 2 DbContext's, one that is more management, and does all the stuff that is behind the ManagementKey checks, and one that is tenant scoped. The management one is the one you use to do migrations and such. The downside is that you have to implement some DbSets multiple times.

abergs: Thanks for all the input @justindbaur, I will experiment a bit and see if I can cook something up that looks OK and then we'll go from there.

src/Api/Helpers/AddDatabaseExtensionMethod.cs
```diff
@@ -16,6 +16,10 @@ public static IServiceCollection AddDatabase(this IServiceCollection services, I
         // Database information
         var sqlite = configuration.GetConnectionString("sqlite");
         var mssql = configuration.GetConnectionString("mssql");
+
+        sqlite = configuration.GetValue<string>("ConnectionStrings.Sqlite.Api", sqlite);
+        mssql = configuration.GetValue<string>("ConnectionStrings.Mssql.Api", mssql);
```
justindbaur, line 21: Same comment as the AdminConsole one.

justindbaur (changes requested):

abergs:
> I would like to work towards only having a single place to define your connection string. I'd like to phase out the old one
@justindbaur agreed. But I wanted to be able to merge and deploy this without a breaking change in prod.

PR #44, opened 2023-06-26
Add CI/CD

joseph-flinn:
## Objective

Add CI/CD to support the Engineering SDLC of Passwordless.dev.

### Versioning Strategy

We are experimenting with automatic version bumping with SemVer. All version bumps default to an increment of the patch version of the latest GitHub tag. If the major or minor version is updated in the `Directory.Build.Props`, the patch version is reset to 0.

### Pushing commits

Every commit that is pushed to GitHub on all branches except for `main` has tests and test builds run to have as short of a feedback loop as possible for the engineer working on the PR.

### Merges to `main`

Every merge to `main` will run through the tests, will be built, and then automatically deployed to our DevTest Environment. This also assists in keeping the feedback loop short and helps the engineer see their code in a full environment as soon as it is accepted to `main`

### Deploys

For now, deploys to any higher environments are manually triggered from GitHub Actions. These deploys will also result in a draft GitHub Release being created with generated Release notes.

### DIFF
--- a/.github/workflows/_deploy.yml
+++ b/.github/workflows/_deploy.yml
@@ -1,0 +1,49 @@
+---
+name: _deploy
+run-name: Deploy to ${{ inputs.environment }}
+
+on:
+  workflow_call:
+    inputs:
+      environment:
+        type: string
+        required: true
+        description: "Environment to deploy to"
+      project-name:
+        type: string
+        required: true
+        description: "Name of the project to deploy"
+      is-dry-run:
+        type: boolean
+        default: true
+      version:
+        type: string
+        required: true
+
+jobs:
+  deploy:
+    name: Deploy ${{ inputs.project-name }} to ${{ inputs.environment }}
+    runs-on: ubuntu-22.04
+    environment: ${{ inputs.environment }}-${{ inputs.project-name }}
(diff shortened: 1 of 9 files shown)


michalchecinski (approved):

.github/workflows/_deploy.yml
```diff
@@ -0,0 +1,49 @@
...
+        if: ${{ ! inputs.is-dry-run }}
+        uses: Azure/login@1f63701bf3e6892515f1b7ce2d2bf1708b46beaf  # v1.4.3
+        with:
+          creds: ${{ secrets.APP_SERVICE_SP }}  # From the environment.secrets
+
+      - name: Configure & Deploy App
+        if: ${{ ! inputs.is-dry-run }}
+        run: |
+          az webapp config appsettings set -g ${{ secrets.AZURE_RESOURCE_GROUP_NAME }} \
+            -n ${{ secrets.AZURE_APP_NAME }} --slot staging \
```
abergs, line 45: @joseph-flinn API and Admin runs on different apps, so we need different variables based on project name

joseph-flinn: These should be handled in the GitHub Environment. The secret name will be the same for each

PR #45, opened 2023-06-29
Breaking change: Uses specific value for db connection string.

abergs:
Two motivations:
* Simplify to only have one config value to decide connection string
* Bug: We didn't properly use our new values

### DIFF
--- a/src/AdminConsole/appsettings.Development.json
+++ b/src/AdminConsole/appsettings.Development.json
@@ -11,7 +11,7 @@
     "CustomerAuthPlanName": "Professional Test"    
   },
   "ConnectionStrings": {
-    "sqlite": "Data Source=adminconsole_dev.db"
+    "sqlite:admin": "Data Source=adminconsole_dev.db"
   },
   "Passwordless": {
     "ApiUrl": "http://localhost:7001",
--- a/src/AdminConsole/appsettings.Development.json
+++ b/src/AdminConsole/appsettings.Development.json
@@ -11,7 +11,9 @@
     "CustomerAuthPlanName": "Professional Test"    
   },
   "ConnectionStrings": {
-    "sqlite:admin": "Data Source=adminconsole_dev.db"
+    "sqlite": {
+      "admin": "Data Source=adminconsole_dev.db"
+    }
   },
   "Passwordless": {
     "ApiUrl": "http://localhost:7001",
--- a/src/Api/Helpers/AddDatabaseExtensionMethod.cs
+++ b/src/Api/Helpers/AddDatabaseExtensionMethod.cs
@@ -21,7 +21,8 @@ public static class AddDatabaseExtensionMethod
         {
             services.AddDbContext<DbTenantContext, SqliteContext>((sp, builder) =>
             {
-                builder.UseSqlite(sqlite);
+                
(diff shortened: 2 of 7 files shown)


abergs:
The need to merge this change is if we think it will be valuable to have different name or if we should let the execution env figure out envrionment variabler conflits, like in: https://github.com/passwordless/passwordless-server/commit/bb5b6e86c36effe039106529d15e3bfeee565b64

src/Api/Helpers/AddDatabaseExtensionMethod.cs
```diff
@@ -14,27 +14,23 @@ public static class AddDatabaseExtensionMethod
...
-        mssql = configuration.GetValue<string>("ConnectionStrings:Mssql:Api", mssql);
+        var sqlite = configuration.GetConnectionString("sqlite:api");
+        var mssql = configuration.GetConnectionString("mssql:api");
 
         if (!string.IsNullOrEmpty(sqlite))
         {
             services.AddDbContext<DbTenantContext, SqliteContext>((sp, builder) =>
             {
-                var config = sp.GetRequiredService<IConfiguration>();
-                builder.UseSqlite(config.GetConnectionString("sqlite"));
```
justindbaur, line 28: It's fine to remove this but I'll explain why I did this and leave the decision up to you. By using `IConfiguration` from the service provider instead of the outer scope one you are able to keep this method `static`. This way it won't capture any outer scope, requires less allocations, and should run a tad faster.

There are no bugs that should arise from capturing `IConfiguration` because it's a singleton, it's lifetime will match that of the Host so it won't unnecessarily be kept from being GC'ed (because it won't ever be), and changes to it aren't done through creating a new reference. So in this instance it's safe but I generally try to not capture scope if I can help it. 

Decision is up to you though.

src/Api/appsettings.Development.json
```diff
@@ -1,6 +1,6 @@
 {
   "ConnectionStrings": {    
-    "sqlite": "Data Source=passwordless_dev.db"
+    "sqlite:api": "Data Source=passwordless_dev.db"
```
justindbaur, line 3: Same thing.
```suggestion
    "sqlite": {
      "api": "Data Source=passwordless_dev.db"
     }
```

src/AdminConsole/appsettings.Development.json
```diff
@@ -11,7 +11,7 @@
     "CustomerAuthPlanName": "Professional Test"    
   },
   "ConnectionStrings": {
-    "sqlite": "Data Source=adminconsole_dev.db"
+    "sqlite:admin": "Data Source=adminconsole_dev.db"
```
justindbaur, line 14: In JSON I vote to represent it with object syntax.
```suggestion
    "sqlite": {
      "admin": "Data Source=adminconsole_dev.db"
    }
```

justindbaur (approved):
Some suggestions and comments but overall a good improvement that has my seal of approval.

src/Api/Helpers/AddDatabaseExtensionMethod.cs
abergs: Aha I see, yeah that's a valid point. Will add it back with a explainer.

PR #50, opened 2023-07-06
Added support for publishing nuget packages on CI

abergs:
We use this to test nuget packets easy


joseph-flinn (commented):
I'd suggest creating another callable workflow following the pattern in `_test.yml` and replacing the testing logic (everything after the `Checkout Repo` step) with the nuget pack and publish. I would then call that workflow in `cd.yml` after the `test` job

joseph-flinn:
> @joseph-flinn That would require that we build the solution multiple times, yes?

Yeah, that would be building the code again (which I think we are currently doing twice already for `ci`...). Optimally, we'd only be building the code once per collection of pipelines and pull those build artifacts to be reused for the entire flow. But we can work on that in the next iteration. For now, this approach gets us where we want to go a bit quicker.

joseph-flinn (changes requested):
We'll need to add `packages: write` permissions in `cd.yml` and `deploy.yml` as well since they both call the `_test.yml` workflow.

PR #53, opened 2023-07-31
Added FluentValidation. Added rule for max length on Nickname. Added …

jrmccannon:
Added FluentValidation. Added max length rule to Nickname on RegistrationCompleteDTO model.

I've worked with FluentValidation before and enjoy it more than DataAnnotations. We could add more to this and have the validation automatically take place through model binding.  I also added an additional exception block to `FriendlyExceptionsMiddleware` to handle the validation exceptions which will provide a list of invalid properties to any API caller.

### DIFF
--- a/src/Api/Endpoints/Register.cs
+++ b/src/Api/Endpoints/Register.cs
@@ -1,4 +1,5 @@
 ﻿using ApiHelpers;
+using FluentValidation;
 using Passwordless.Api.Authorization;
 using Passwordless.Service;
 using Passwordless.Service.Models;
@@ -31,8 +32,15 @@ public static class RegisterEndpoints
             .RequireCors("default")
             .WithMetadata(new HttpMethodMetadata(new string[] { "POST" }, acceptCorsPreflight: true));
 
-        app.MapPost("/register/complete", async (RegistrationCompleteDTO payload, HttpRequest request, IFido2ServiceFactory fido2ServiceFactory) =>
+        app.MapPost("/register/complete", async (RegistrationCompleteDTO payload, HttpRequest request, IFido2ServiceFactory fido2ServiceFactory, IValidator<RegistrationCompleteDTO> validator) =>
         {
+            var validation = await validator.ValidateAsync(payload);
+
+            if (!validation.IsValid)
+            {
+                throw new ValidationException(validation.Errors);
+            }
+
             var fido2Service = await fido2ServiceFactory.CreateAsync();
             var (deviceInfo, country) = Helpers.GetDeviceInfo(request);
             var result = await fido2Service.RegisterComplete(payload, deviceInfo, country);
--- a/src/Api/Helpers/FriendlyExceptionsMiddleware.cs
+++ b/src/Api/Helpers/FriendlyExceptionsMiddleware.cs
@@ -1,4 +1,5 @@
 using Datadog.Trace;
+using FluentValidation;
 using Microsoft.Data.Sqlite;
 using Passwordless.Service.Helpers;
 
@@ -48,6 +49,21 @@ public class FriendlyExceptionsMiddleware
                 extensions: extras,
                 type: $"https://docs.passwordless.dev/guide/errors.html#{apiException.ErrorCode}");
             await problem.ExecuteAsync(context);
+        }
+        catch (ValidationException validationException)
+        {
+            Tracer.Instance.ActiveScope?.Span.SetException(validationException);
+            _logger.UncaughtException(validationException);
+
+            await Results.ValidationProblem(
+                validationException.Errors
+                    .GroupBy(x => x.PropertyName)
+                    .ToDictionary(x => x.Key, x => x.Select(e => e.ErrorMessage).ToArray()),
+                validationException.Message,
+                validationException.Source,
+                StatusCodes.Status400BadRequest,
+                "One or more validation errors occured"
+            ).ExecuteAsync(context);
         }
         catch (Exception exception)
         {
--- a/src/Api/Api.csproj
+++ b/src/Api/Api.csproj
@@ -17,6 +17,8 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Datadog.Trace" Version="2.30.0" />
+    <PackageReference Include="FluentValidation" Version="11.6.0" />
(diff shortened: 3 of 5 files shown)


abergs:
Thank you Jared! I don't have much experience with `FluentValidation`. Added @justindbaur to align with expertise from things we do in the other code bases as well to review the ProblemDetails stuff since he worked on it :) 

One thing that DataAnnotations gives us is that EntityFramework will apply the same values to the database, so 
tweaking the model to have the MaxLength attributes gives us quite a lot of value from very little input. 

```chsarp
[MaxLength(255)]
public string Nickname { get; set; }
```

From doing a quick review on the codebase, I can see that the API has very little existing `structured` validation (we should add it to make sure to goes down to the db) so it's a great time to decide either for or against FlutValidation. 

AdminConsole for comparison uses DataAnnotations.

Can I hear some more debate, pro/cons?

src/Api/Endpoints/Register.cs
```diff
@@ -31,8 +32,15 @@ public static void MapRegisterEndpoints(this WebApplication app)
...
             .WithMetadata(new HttpMethodMetadata(new string[] { "POST" }, acceptCorsPreflight: true));
 
-        app.MapPost("/register/complete", async (RegistrationCompleteDTO payload, HttpRequest request, IFido2ServiceFactory fido2ServiceFactory) =>
+        app.MapPost("/register/complete", async (RegistrationCompleteDTO payload, HttpRequest request, IFido2ServiceFactory fido2ServiceFactory, IValidator<RegistrationCompleteDTO> validator) =>
         {
+            var validation = await validator.ValidateAsync(payload);
+
+            if (!validation.IsValid)
+            {
+                throw new ValidationException(validation.Errors);
```
justindbaur, line 41: I'd rather just return `return ValidationProblems` from here instead of throwing.

src/Api/Helpers/FriendlyExceptionsMiddleware.cs
```diff
@@ -49,6 +50,21 @@ private async Task SafeNext(HttpContext context)
...
+            await Results.ValidationProblem(
+                validationException.Errors
+                    .GroupBy(x => x.PropertyName)
+                    .ToDictionary(x => x.Key, x => x.Select(e => e.ErrorMessage).ToArray()),
+                validationException.Message,
+                validationException.Source,
+                StatusCodes.Status400BadRequest,
+                "One or more validation errors occured"
+            ).ExecuteAsync(context);
+        }
```
justindbaur, line 67: I'm fine with this generally but only for exceptions thrown outside of the endpoint context, like if we have to do more validation inside a service then we can throw and have them caught here. But if we have the validation problems inside the endpoint we can just return them from there and avoid throwing an exception.

justindbaur (commented):
@abergs You can't _just_ add the data annotations to get validation in minimal APIs, that was left out and you have to bring your own. 

The other player I know of in this space is [`MiniValidation`](https://github.com/DamianEdwards/MiniValidation) which does utilize the data annotations but I generally like FluentValidation more since it's a bit more strictly typed. With data annotations you can do things like add `[MaxLength]` on an object property and it doesn't do anything. 

I don't love them all being registered via reflection over the assembly but that is probably because I care about cold start performance and possible uses of NativeAOT down the road. So I can leave that up to you, because it is a good way of making sure you don't forget to add one.

The one thing I _really_ dislike is the code formatting of their codebase but we don't have to look at that 😆.

src/Api/Helpers/FriendlyExceptionsMiddleware.cs
jrmccannon: Thank you. I saw the other two exceptions and didn't know if this was the pattern we wanted.  I've removed the validation block.

src/Api/Endpoints/Register.cs
jrmccannon: Made the change.  I didn't know what to do with the `detail` or `instance` parameters. So I left them `null` for now since that's the default.

jrmccannon:
@abergs 
Another benefit of FluentValidation is that it allows for robust and complex validation rules against objects. `DataAnnotations` can only handle simple rules and anything more complicated you'll have to create yourself.  It doesn't have good building blocks for more complicated rule sets.

As far as adding restrictions to the EF contexts, I'd rather we did it through the context definition using its FluentAPI. 

@justindbaur 
I'd honestly rather everything be as explicit as possible, so I've added a `ServiceCollection` extension method for adding the `AbstractValidator<T>` classes.

justindbaur:
I also like the fluent API for EF Core configuration, for pretty much the same reason as @jrmccannon mentions for validation. You can only get so far with data annotation so you will be forced to do more elsewhere. With EF that is adding it into the context and now you have two places configuration comes from, but if you go all in with the fluent way, you only have it in one place.

jrmccannon:
To summarize discussions around adding `FluentValidation` to the code base.

`FluentValidation` gives us a quick, ready-made way to build model validation rules into the application.  The syntax and pattern is easy to follow and implement. It standardizes all validation messages around which properties are invalid and why.  We can do manual validation to call the `Validate` method off the `IValidator<T>` interface.  It also opens up the opportunity for adding the validation step into the `.NET` request pipeline to automatically do validations when they exist.  `FluentValidation` has `.NET Standard 2.0` support which means it will work with `.NET Framework 4.6.1`.

The downside is that the `AbstractValidator<T>` classes are a bit hidden and disconnected from the actual object being validated.  The more those classes are abstracted away from the code, the more convoluted the code flow is.  It also means the application takes on `FluentValidation` and its dependencies.  

If the decision is to use `FluentValidation`, we should make it consistent across the solution.

jrmccannon:
We should definitely use FluentValidation for all entry points into the application.  That way, any invalid requests will return a uniform error message listing what's wrong with the request. 

Reading the documentation, we can use it to drive the Razor page validation as well.

abergs (commented):
Awaiting some removals

jrmccannon:
Removed FluentValidation at this time. I was having trouble getting access to the validation rules to be used in Html5Validation.cs (adding required attribute on input fields to match form models).  We do not need the complex rules quite yet so I went back to DataAnnotations and using MiniValidation to validate request models for the Api application.

abergs (approved):

PR #60, opened 2023-08-04
Implement background service to delete applications that are schedule…

jonashendrickx:
…d for deletion.


src/AdminConsole/ApiHelper.cs
```diff
@@ -43,4 +43,26 @@ public async Task<NewAppResponse> CreateApplication(NewAppOptions registerOption
         res.EnsureSuccessStatusCode();
         return await res.Content.ReadFromJsonAsync<NewAppResponse>();
     }
+
+    public async Task<IEnumerable<string>> ListApplicationsPendingDeletionAsync()
+    {
+        try
+        {
+            var res = await _client.GetAsync("apps/list_pending_deletion");
```
abergs, line 51: I think dashes are `-` preferred by the industry? But I don't have a strong opinion, `_` is fine for me.

src/Service/Storage/Ef/DbTenantContext.cs
```diff
@@ -48,37 +48,41 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
                     v => string.Join(',', v),
                     v => v.Split(',', StringSplitOptions.RemoveEmptyEntries));
         });
 
-        var ap = modelBuilder.Entity<AliasPointer>()
-            .HasQueryFilter(c => c.Tenant == Tenant)
+        modelBuilder.Entity<AliasPointer>()
             .HasKey(x => new { x.Tenant, x.Alias });
 
+        if (Tenant != null)
```
abergs, line 77: @jonashendrickx I'm a bit hesitant to this change. We want this code path to be very defensive. If we make this change with`if (Tenant != null)` we should make other defensive measures in the TenantProvider to throw exception if value is null, or only allowing a null value if the ITenantProvider returns true for property `AllowNull` or `NonScoped`etc.

abergs (changes requested):

PR #66, opened 2023-08-11
Delete an organization

jonashendrickx:


abergs (commented):
@jonashendrickx We should also delete the user account as part of this process I think, since the account is not usable without a organization.

PR #67, opened 2023-08-12
PAS90: Users List should have empty value for column LastSeenAt

jonashendrickx:
- Users list should have empty value for column 'Last Seen At' or 'Last Used at' instead of showing DateTime.MinValue
- Bumped NuGet version to 0.0.13 ([requires PR to be approved and NuGet to be published](https://github.com/passwordless/passwordless-dotnet/pull/10))


jonashendrickx:
Screenshot 2023-08-12 at 10 14 42

abergs (approved):

PR #80, opened 2023-08-25
PAS164: Make it easier to spot when an app is marked for deletion.

jonashendrickx:


src/AdminConsole/Pages/App/Settings/Settings.cshtml
```diff
@@ -21,7 +21,7 @@
 @if (Model.PendingDelete)
 {
     <form method="post" asp-page-handler="Cancel">
-        <p>@Model.ApplicationId will be deleted on <local-time datetime="@Model.DeleteAt?.ToString("u")"></local-time>. If you would like to cancel this action, click "Cancel".</p>
+        @Html.Partial("Shared/Alerts/_DangerAlert", $"{@Model.ApplicationId} will be deleted on \"{Model.DeleteAt?.ToString("g")} GMT\". If you would like to cancel this action, click \"Cancel\".")
```
abergs, line 24: I can't wait for razor components. Will make these things a lot easier.

abergs: While they are the same, would it make sense to keep to UTC? I feel like that is more commonly known by non-eu developers.

abergs (approved):

abergs:
@jonashendrickx LGTM :shipit:

PR #88, opened 2023-09-07
Simplified layering in Adminconsole

abergs:


src/AdminConsole/CurrentContextMiddleware.cs
```diff
@@ -116,9 +118,13 @@ public CurrentContextMiddleware(RequestDelegate next)
...
         var hasOrgIdClaim = httpContext.User.HasClaim(x => x.Type == "OrgId");
 
-        return hasOrgIdClaim
+        var action = hasOrgIdClaim
             ? InvokeCoreAsync(httpContext, appId, currentContext, dbContext, passwordlessClient, options.Value)
             : _next(httpContext);
+
+        await action;
+
+        await auditLogger.FlushAsync();
```
abergs, line 127: Admittedly, this is not beutiful.

jrmccannon: We can move this to a new middleware

jrmccannon (approved):

PR #94, opened 2023-09-17
PAS37: Show error message when WebAuthn is not supported.

jonashendrickx:
- Showing error message when WebAuthn isn't supported in browsers like Tor Browser.
- Refactoring to reference Passwordless Client Js library from a partial view.


jonashendrickx:
image

image

image

PR #95, opened 2023-09-18
PAS106: Fix validation for special characters

jonashendrickx:


src/AdminConsole/Pages/Organization/Settings.cshtml.cs
```diff
@@ -70,5 +76,32 @@ public async Task<IActionResult> OnPostDeleteAsync()
...
+    /// <summary>
+    /// Whether the organization can be deleted.
+    /// </summary>
+    public bool CanDelete => ApplicationsCount == 0;
+
+    /// <summary>
+    /// The amount of active applications belonging to the organization.
+    /// </summary>
+    [BindProperty(SupportsGet = true)]
+    public int ApplicationsCount { get; set; }
```
abergs, line 106: @jonashendrickx Is there a reason why we allow these parameters to be bound? This allows someone to setup through querystrings.

abergs (changes requested):

src/AdminConsole/Pages/Organization/Settings.cshtml.cs
```diff
@@ -61,7 +67,7 @@ public async Task<IActionResult> OnPostDeleteAsync()
             }
         }
 
-        var isDeleted = await _dataService.DeleteOrganizationAsync(Organization.Id);
+        var isDeleted = await _dataService.DeleteOrganizationAsync(Id);
```
abergs, line 70: I'm also wondering why the Delete call needs a ID at all, since it should probably used the one it has internally?

jonashendrickx: I can refactor that bit separately

src/AdminConsole/Pages/Organization/Settings.cshtml.cs
jonashendrickx: Just a leftover mistake.

src/AdminConsole/Pages/Organization/Settings.cshtml.cs
```diff
@@ -38,12 +35,27 @@ public class SettingsModel : PageModel
...
     {
-        Organization = await _dataService.GetOrganizationWithData();
+        await LoadData();
+    }
+
+    private async Task LoadData()
+    {
+        var organization = await _dataService.GetOrganizationWithData();
+        Name = organization.Name;
+        ApplicationsCount = organization.Applications.Count;
```
Tyrrrz, line 45: Why not just set `this.Organization = organization` instead of splitting into separate props? To reduce the amount of data transferred?

Tyrrrz (approved):

src/AdminConsole/Pages/Organization/Settings.cshtml.cs
abergs: @Tyrrrz It wouldn't be possible to use Organization as a property (only because we refactored a bit in the last commit!) -  and I think in this case we might have been able to. Otherwise I think it's a good idea to try to be explicit (where it's practical) with what objects we expose to the view and/or modelbinding to add friction to someone accidently adding binding on a sensitive object.

abergs (approved):

PR #96, opened 2023-09-18
PAS64: Limit users to 10k on free plan.

jonashendrickx:
Goes with https://github.com/passwordless/docs/pull/44


jonashendrickx:
Scripts have not been generated yet for the in-house migration.

jonashendrickx:
image

We only needed to show the error in the playground. But error is returned by the API when a new user (with new user id) tries to register.

jonashendrickx:
Configuration is set in appsettings.json. No extra configuration required on server.

abergs:
@jonashendrickx Let's have a call tomorrow and walk through some of these changes!

PR #97, opened 2023-09-19
Add `.idea` to `.gitignore`

Tyrrrz:
Contains user-specific files that shouldn't be added to version control


jonashendrickx (approved):

Berezhnyk:
I don't know if the whole directory should be added. The `.idea` might contain team-shared settings. 
Article for the reference https://intellij-support.jetbrains.com/hc/en-us/community/posts/360002772419-Best-way-to-share-settings-in-a-team-and-allow-individual-ones-

Tyrrrz:
> I don't know if the whole directory should be added. The `.idea` might contain team-shared settings. Article for the reference [intellij-support.jetbrains.com/hc/en-us/community/posts/360002772419-Best-way-to-share-settings-in-a-team-and-allow-individual-ones-](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360002772419-Best-way-to-share-settings-in-a-team-and-allow-individual-ones-)

Indeed, but I don't think it was intentional in this case.

PR #113, opened 2023-09-28
PAS152: Prepare to migrate to Fido2 4.0.0.-beta4

jonashendrickx:


tests/Api.IntegrationTests/CredentialsTest.cs
```diff
@@ -24,18 +25,20 @@ public async Task CredentialsList()
         context.Credentials.Add(new EFStoredCredential()
         {
             DescriptorId = "test"u8.ToArray(),
+            DescriptorType = PublicKeyCredentialType.PublicKey,
             UserHandle = "1"u8.ToArray(),
             Tenant = "test",
         });
 
-        await context.SaveChangesAsync();
+        var rows = await context.SaveChangesAsync();
```
abergs, line 33: Are we using rows?

abergs (commented):
We'd also want to enable the rk credProps when discoverable is true.

src/Service/Fido2ServiceEndpoints.cs
```diff
@@ -316,37 +312,38 @@ public async Task<TokenResponse> SignInComplete(SignInCompleteDTO request, strin
...
 
-        var userId = Encoding.UTF8.GetString(creds.UserHandle);
+        var userId = Encoding.UTF8.GetString(credential.UserHandle);
 
         var tokenData = new VerifySignInToken
         {
             UserId = userId,
             Success = true,
             Origin = request.Origin,
-            RPID = request.RPID,
```
jonashendrickx, line 344: @abergs Here

PR #120, opened 2023-09-29
Table css refactoring & tag helpers fix for Tailwind CSS

jonashendrickx:


abergs:
On QA:
CleanShot 2023-09-29 at 16 42 19

ThisPR:
CleanShot 2023-09-29 at 16 42 42

Color on links is gone and indent is missaligned on row vs header

jonashendrickx:
> On QA: CleanShot 2023-09-29 at 16 42 19
> 
> ThisPR: CleanShot 2023-09-29 at 16 42 42
> 
> Color on links is gone and indent is missaligned on row vs header

Fixed!

Tyrrrz:
For what's it worth Tailwind docs recommend _against_ using `@apply` together with composite CSS styles to manage duplication: https://tailwindcss.com/docs/reusing-styles#compared-to-css-abstractions

Instead they recommend favoring partials or components. I'm not sure what Razor's capabilities are in this regard, but maybe we can use partials instead? E.g. `<partial name="_PrimaryButton" />` etc.

jonashendrickx:
> For what's it worth Tailwind docs recommend _against_ using `@apply` together with composite CSS styles to manage duplication: https://tailwindcss.com/docs/reusing-styles#compared-to-css-abstractions
> 
> Instead they recommend favoring partials or components. I'm not sure what Razor's capabilities are in this regard, but maybe we can use partials instead? E.g. `<partial name="_PrimaryButton" />` etc.

It actually recommends using '@apply' for creating reusable components. We will also be migrating away from Razor Pages likely with .NET 8 to something that's more component-friendly.

> For what's it worth Tailwind docs recommend _against_ using `@apply` together with composite CSS styles to manage duplication: https://tailwindcss.com/docs/reusing-styles#compared-to-css-abstractions
> 
> Instead they recommend favoring partials or components. I'm not sure what Razor's capabilities are in this regard, but maybe we can use partials instead? E.g. `<partial name="_PrimaryButton" />` etc.

Down at the bottom that statement is contradicted again. And it feels expensive to create partials for just simple things as buttons. Because partials are at least 5 times slower than tag helpers, and even more compared to native HTML elements.

It does help for creating reusable components. Partials are also at least 5 times slower compared to TagHelpers, so you can imaging they're a magnitude slower than using native HTML elements.

We might migrate away from Razor Pages to something else with .NET 8.

Tyrrrz:
> It actually recommends using '@apply' for creating reusable components.

Indeed, but seemingly only as a last resort when no other component system is available.

> And it feels expensive to create partials for just simple things as buttons. Because partials are at least 5 times slower than tag helpers

Isn't it the same as a component in other frameworks, basically? In which case, small simple things get extracted to components all the time.

The fact that partials are slow is unfortunate, but does it have a measurable impact in terms of all other performance overhead we have (not withstanding network latency and SSR)?

In any case, I'm okay with us using CSS for reusability for the time being, but we should definitely revisit this approach in .NET 8 where we'll have Blazor Components available as an option (from what I understand, right?).

jonashendrickx:
> > It actually recommends using '@apply' for creating reusable components.
> 
> Indeed, but seemingly only as a last resort when no other component system is available.
> 
> > And it feels expensive to create partials for just simple things as buttons. Because partials are at least 5 times slower than tag helpers
> 
> Isn't it the same as a component in other frameworks, basically? In which case, small simple things get extracted to components all the time.
> 
> The fact that partials are slow is unfortunate, but does it have a measurable impact in terms of all other performance overhead we have (not withstanding network latency and SSR)?
> 
> In any case, I'm okay with us using CSS for reusability for the time being, but we should definitely revisit this approach in .NET 8 where we'll have Blazor Components available as an option (from what I understand, right?).

That's the goal yes! I just wanted to make the code more maintainable in the meantime. Rome wasn't built in one day either. :-)

abergs:
Yep, let's definitely revisit later when .net 8 is out.

Ideally I would like to land somewhere in between tailwind and bootstrap.

PR #122, opened 2023-10-02
Improve Entity Framework readme

abergs:


jonashendrickx (approved):

Tyrrrz:
May be obsoleted by #128

abergs:
@jrmccannon @Tyrrrz Just did a brief look at #128 which seems to contain some nice documentation, but doesn't mention that we want to run idempotent script generation for prod (which we do). Bring in whatever you find useful from this readme PR to #128 .

PR #123, opened 2023-10-02
Adding Reload and Redirect to Organization/Verify

jrmccannon:
This will add a 15 second reload to the /Organization/Verify page.

After the user has signed up for an account, we send them to /organization/verify/.  Here we will now refresh the page every 15 seconds and if they've clicked the link or are verified, it will redirect them to the /account/useronboarding page to set up their passkey.

### DIFF
--- a/src/AdminConsole/Pages/Organization/Verify.cshtml
+++ b/src/AdminConsole/Pages/Organization/Verify.cshtml
@@ -7,8 +7,8 @@
 }
 
 @section Head {
-    @* This to refresh the page to check if the user has verified and logged in *@
-    <meta http-equiv="refresh" content="15"/>
+    // This to refresh the page to check if the user has verified and logged in
+    <meta http-equiv="refresh" content="15" />
 }
 
 <p>We've sent a confirmation link to the email address you provided. Click on the link in the email to verify your account</p>
--- a/src/AdminConsole/Pages/Organization/Verify.cshtml
+++ b/src/AdminConsole/Pages/Organization/Verify.cshtml
@@ -8,6 +8,8 @@
 
 <p>We've sent a confirmation link to the email address you provided. Click on the link in the email to verify your account</p>
 
+<meta http-equiv="refresh" content="10" />
+
 <environment names="Development">
     @{
         var mailPath = System.IO.Path.GetFullPath("mail.md");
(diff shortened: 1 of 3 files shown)


Tyrrrz:
> This will add a 15 second reload to the /Organization/Verify page.

image

Lies. Lies everywhere.

src/AdminConsole/Pages/Organization/Verify.cshtml
```diff
@@ -8,6 +8,8 @@
 
 <p>We've sent a confirmation link to the email address you provided. Click on the link in the email to verify your account</p>
 
+<meta http-equiv="refresh" content="10" />
```
Tyrrrz, line 11: Maybe add a  to explain why we need this?

Tyrrrz (approved):

src/AdminConsole/Pages/Organization/Verify.cshtml
jonashendrickx, line 11:
```suggestion
@section Head {
    <meta http-equiv="refresh" content="15" />
}
```

Render it inside the <head> tag.

Then in _Layout.cshtml render the section:

```html
<head>
    **@await RenderSectionAsync("Head", false)**
</head>
```

jonashendrickx (changes requested):

jonashendrickx (approved):

src/AdminConsole/Pages/Organization/Verify.cshtml
```diff
@@ -6,16 +6,19 @@
     ViewData["Title"] = "Verify your account";
 }
 
-<p>We've sent a confirmation link to the email address you provided. Click on the link in the email to verify your account</p>
+@section Head {
+    <!-- This to refresh the page to check if the user has verified and logged in -->
```
Tyrrrz, line 10: You could use a C# comment (`//`) to make this invisible to the users. In some places you need to use (`@* ... *@`) for server-side comments, but since it's in a section, `//` should work.

src/AdminConsole/Pages/Organization/Verify.cshtml
```diff
@@ -6,16 +6,19 @@
...
 
-<meta http-equiv="refresh" content="10" />
+<p>We've sent a confirmation link to the email address you provided. Click on the link in the email to verify your account</p>
 
 <environment names="Development">
     @{
         var mailPath = System.IO.Path.GetFullPath("mail.md");
     }
-    <br />
+    <br/>
```
Tyrrrz, line 20: Nit: `<br>` is a void element, so it doesn't need the `/` -- `<br>` should work. We can establish dominance over whoever decides to read this code by highlighting that we have read the W3C specification on HTML tags.

PR #124, opened 2023-10-03
Fix scheduled application deletion

jrmccannon:
Currently, `ApplicationDeletionBackgroundService` was calling the wrong endpoint to delete the application in API. I fixed the url and also added a check in to provide additional logging when an application deletion failed during the API call.

### DIFF
--- a/src/AdminConsole/Services/ApplicationDeletionBackgroundService.cs
+++ b/src/AdminConsole/Services/ApplicationDeletionBackgroundService.cs
@@ -1,6 +1,5 @@
 using AdminConsole.Db;
 using AdminConsole.Models;
-using Microsoft.EntityFrameworkCore;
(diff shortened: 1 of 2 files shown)


abergs (commented):
Were you able to verify if they got deleted locally?

jrmccannon:
Yes. I was able to see the background service hit api on the correct url and the application was deleted on both API and AdminConsole. 

The `if` check I put in is pretty safe because we call out to API to gather all pending deletion applications.  So if it wasn't deleted in API we should fail to delete it in admin console to keep them in sync. We'd need another type of solution to get everything that was successfully deleted in API but lingers in Admin.  However, future database plans may solve this for us.

jonashendrickx (approved):

PR #138, opened 2023-10-12
Migrate to use IDbContextFactory to allow pooling of our connections.

jonashendrickx:


src/AdminConsole/Db/AddDatabaseExtensionMethod.cs
```diff
@@ -12,26 +18,43 @@ public static IServiceCollection AddDatabase(this IServiceCollection services, W
         var migrating = builder.Configuration.GetValue<string>("ef_migrate");
         if (migrating == "1")
         {
-            services.AddDbContext<MssqlConsoleDbContext>();
-            services.AddDbContext<SqliteConsoleDbContext>();
+            services.AddDbContextFactory<MssqlConsoleDbContext>();
+            services.AddDbContextFactory<SqliteConsoleDbContext>();
```
Tyrrrz, line 22: Question: why are we not using `AddDatabaseContext(...)` (your method) here too?

src/AdminConsole/EventLog/Loggers/EventLoggerEfReadStorage.cs
```diff
@@ -4,26 +4,34 @@
 
 namespace Passwordless.AdminConsole.EventLog.Loggers;
 
-public class EventLoggerEfReadStorage : IEventLoggerStorage
+public class EventLoggerEfReadStorage<TDbContext> : IEventLoggerStorage where TDbContext : ConsoleDbContext
```
Tyrrrz, line 7: Why do we need this change? Couldn't we still do `IDbContextFactory<ConsoleDbContext>` if we registered the corresponding abstraction?

src/AdminConsole/Helpers/ClaimsExtension.cs
```diff
@@ -4,9 +4,11 @@ namespace Passwordless.AdminConsole.Helpers;
 
 public static class ClaimsExtension
 {
-    public static int GetOrgId(this ClaimsPrincipal user)
+    public static int? GetOrgId(this ClaimsPrincipal user)
     {
-        int orgId = int.Parse(user.FindFirstValue("OrgId"));
+        var orgIdStr = user.FindFirstValue("OrgId");
+        if (orgIdStr == null) return null;
+        int orgId = int.Parse(orgIdStr);
```
Tyrrrz, line 11: Provide culture-invariant parsing here?

Tyrrrz, line 11:
```suggestion
        var orgId = int.Parse(orgIdStr, CultureInfo.InvariantCulture);
```

src/AdminConsole/Services/ISharedBillingService.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Passwordless.AdminConsole.Services;
+
+public interface ISharedBillingService
+{
+    Task UpdateUsage();
+
+    Task UpdateStripe(string subscriptionItemId, int users);
+
+    Task ConvertFromFreeToPaidAsync(string customerId, string clientReferenceId, string subscriptionId);
```
Tyrrrz, line 11: Can we standardize on the `Async` suffix? Either use it or not use it

Tyrrrz (approved):

src/AdminConsole/Db/AddDatabaseExtensionMethod.cs
jonashendrickx: Because then it would resolve DataService<SqliteConsoleDbContext> and DataService<MsSqlConsoleDbContext> to IDataService. That would not go well.

I don't think this part is even still being used.

src/AdminConsole/EventLog/Loggers/EventLoggerEfReadStorage.cs
jonashendrickx: No because it cannot resolve ConsoleDbContext. We tried to use AddTransient<ConsoleDbContext, SqliteDbContext>() for example. But it still won't work.

With AddDbContext you can do AddDbContext<ConsoleDbContext, SqliteDbContext>(). But not when you use the factory.

src/AdminConsole/Services/ISharedBillingService.cs
jonashendrickx: Yeah agreed. I just thought it might be too much at once.

src/AdminConsole/EventLog/Loggers/EventLoggerEfReadStorage.cs
Tyrrrz: Ok, makes sense

PR #139, opened 2023-10-12
Refactor event logging functions

Tyrrrz:
Removed `EventFunctions` classes and replaced them with extension methods on `IEventLogger`s.


src/AdminConsole/EventLog/DTOs/OrganizationEventDto.cs
```diff
@@ -9,4 +10,27 @@ public record OrganizationEventDto(
...
+        };
+
+    public EventLogEvent ToResponse() =>
+        new(PerformedAt,
+            EventType.ToString(),
+            Message,
+            Severity.ToString(),
+            PerformedBy,
+            Subject,
+            string.Empty);
```
Tyrrrz, line 35: Transformed these from extension methods to member methods because it didn't make much sense to separate them

jonashendrickx (approved):

abergs:
Going further, A micro optimization would to move these to be part of the IEventlogger interface. It would allow them to be no-op when we're using the noop Eventlogger (which will be used in 9/10 cases). (Justin's idea originally)

jrmccannon (approved):

PR #143, opened 2023-10-13
Consolidate and simplify workflows

Tyrrrz:
This is my attempt to simplify the workflows, reduce the number of `yml` files, and to make things more consistent with `passwordless-dotnet`. Also some `csproj` cleanup is included here too.

This is a proof of concept and I'm sure there will be input on some things.


.github/workflows/bump.yml
```diff
@@ -0,0 +1,90 @@
+name: bump
```
Tyrrrz, line 1: Same bump workflow as in the other repo

.github/workflows/main.yml
```diff
@@ -0,0 +1,253 @@
...
+          dotnet test
+          --no-restore
+          --no-build
+          --configuration Release
+          --logger "trx;LogFileName=pw-test-results.trx"
+
+      - name: Report test results
+        uses: dorny/test-reporter@c9b3d0e2bd2a4e96aaf424dbaa31c46b42318226  # v1.6.0
+        # Only collect test results if the tests actually ran
+        if: ${{ steps.tests.outcome != 'cancelled' && steps.tests.outcome != 'skipped' }}
```
Tyrrrz, line 91: This should prevent this step from failing when the tests step hasn't even been run (i.e. build or restore fails)

.github/workflows/main.yml
```diff
@@ -0,0 +1,253 @@
...
+          suffix="ci-$time-$ref_clean-${{ github.run_id }}"
+          echo "version=0.0.0-$suffix" >> $GITHUB_OUTPUT
+
+      - name: Publish API
+        run: >
+          dotnet publish src/Api/
+          --no-restore
+          --no-build
+          --configuration Release
+          --runtime win-x86
```
Tyrrrz, line 140: I took this from the existing workflow, but why are we targeting this runtime specifically?

.github/workflows/main.yml
```diff
@@ -0,0 +1,253 @@
...
+      # for the apps, so as to clearly indicate that it's not a stable version release.
+      - name: Generate prerelease version
+        id: prerelease-version
+        if: ${{ github.event_name != 'release' }}
+        run: |
+          time=$(date +%s)
+          ref="${{ github.head_ref || github.ref_name }}"
+          ref_clean="${ref/\//-}"
+          suffix="ci-$time-$ref_clean-${{ github.run_id }}"
+          echo "version=0.0.0-$suffix" >> $GITHUB_OUTPUT
```
Tyrrrz, line 132: Same CI versioning approach as in the other repo, but different from what it was before. Open for input.

.github/workflows/main.yml
```diff
@@ -0,0 +1,253 @@
...
+  # Deploy the apps
+  deploy:
+    needs:
+      # Technically, it's not required for the format job to succeed for us to push the package,
+      # so we may consider removing it as a prerequisite here.
+      - format
+      - test
+      - pack
+    runs-on: ubuntu-latest
+    environment: devtest
```
Tyrrrz, line 179: One environment for both projects is used here. Currently, we have one for each project. Not sure if there's a lot of reason for that, given that (from what I understand) both API and AdminConsole are deployed in tandem. Open for input.

.github/workflows/main.yml
```diff
@@ -0,0 +1,253 @@
...
+          creds: ${{ secrets.APP_SERVICE_SP }}
+
+      - name: Configure app settings
+        # Temporary
+        if: ${{ false }}
+        run: >
+          az webapp config appsettings set
+          --resource-group ${{ secrets.AZURE_RESOURCE_GROUP_NAME }}
+          --name ${{ secrets.AZURE_APP_NAME }}
+          --slot staging
```
Tyrrrz, line 213: `DD_VERSION` isn't being set here (it's a bit more effort to retrieve the actual version in this setup). Do we want to continue setting it like this or find a better way (i.e. resolve in code instead)?

.github/workflows/main.yml
```diff
@@ -0,0 +1,253 @@
...
+
+      - name: Push AdminConsole
+        # Temporary
+        if: ${{ false }}
+        run: >
+          az webapp deployment source config-zip
+          --resource-group ${{ secrets.AZURE_RESOURCE_GROUP_NAME }}
+          --name ${{ secrets.AZURE_APP_NAME }}
+          --slot staging
+          --src AdminConsole.zip
```
Tyrrrz, line 234: This currently deploys to staging, but we can also deploy to production. If we're worried about leaking secrets, we can have another job in a private repo and run it here via a secret webhook (that way it's still triggered by this workflow).

Directory.Build.props
```diff
@@ -1,9 +1,17 @@
 <Project>
   <PropertyGroup>
     <TargetFramework>net7.0</TargetFramework>
-    <BaseVersion>1.0</BaseVersion>
+    <Version>1.0</Version>
```
Tyrrrz, line 4: This will probably be need to set to whatever is the actual version currently

Tyrrrz:
Interestingly, the tests are failing with this workflow with exceptions indicating at issues with the SQLite schema? I'm confused

.github/workflows/main.yml
abergs: @Tyrrrz It's what is used in our prod env. However, not sure pro/cons are for not mentioning it.

.github/workflows/main.yml
abergs: It's fine. However, I think the secrets may differ between API/AdminConsole, so we would need to prefix their name with api/adminconsole

.github/workflows/main.yml
abergs: I'd love to find a better way - I have a ticket open with DataDog to figure it out, but not sure if possible

.github/workflows/main.yml
abergs: I think another repo is a good approach for now. I can set one up.

.github/workflows/main.yml
abergs: We would probably need to still set it here, will update if I hear another solution

abergs:
@Tyrrrz That is indeed very weird. I don't have the faintest.

.github/workflows/main.yml
Tyrrrz: Worst case, I'd set the variable on program startup since it's easier to retrieve version there.

Tyrrrz:
@abergs I've fixed most issues. Currently, it looks like `az login` fails because the credentials are not provided in secrets. It might be because the `environment` has changed? 🤔

abergs:
@Tyrrrz Most definitely.
Will set secrets to devtest.

.github/workflows/main.yml
Tyrrrz: Looks like you're already setting the version in code here:

image

abergs:
@Tyrrrz As long as we're tracking the issues with DD_VERSION I'm OK with merging this.

abergs (approved):

PR #145, opened 2023-10-16
Dates Show as "Invalid Date"

jrmccannon:
When removing `.ToLocalTime()` from razor pages, some mac browsers stopped being able to process the datetime string when calling `new Date()`.  Added `.ToString("O")` so it would format to ISO-8601 standard.

Also added `<local-time />` component to Log pages.

### DIFF
--- a/src/AdminConsole/TagHelpers/LocalTime.cs
+++ b/src/AdminConsole/TagHelpers/LocalTime.cs
@@ -1,0 +1,15 @@
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace Passwordless.AdminConsole.TagHelpers;
+
+[HtmlTargetElement("local-time")]
+public class LocalTime : TagHelper
+{
+    [HtmlAttributeName("datetime")]
+    public DateTime DateTime { get; set; }
+
+    public override void Process(TagHelperContext context, TagHelperOutput output)
+    {
+        output.Attributes.Add("datetime", DateTime.ToString("O"));
+    }
+}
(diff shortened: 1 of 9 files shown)


jonashendrickx (changes requested):
If you want to pass it in ISO 8601 format, maybe wrap it in a tag helper so we can do something as:

<local-time datetime="@user.LastUsedAt.Value"></local-time>

The current implementation is done in Javascript I believe, it would be nice if you can migrate it. There's also an example on how you can test the rendering with unit tests.

jrmccannon:
> The current implementation is done in Javascript I believe, it would be nice if you can migrate it. There's also an example on how you can test the rendering with unit tests.
The reason its done in javascript is because Razor is rendered on the server and we're using the Javascript Date object to convert from `UTC` to the client's local time zone.  From what I've seen, Razor and Blazor both require javascript to access the client's local time.

> Then we can avoid having .ToString("O") all over our code.
I added a TagHelper to convert the DateTime to the ISO 8601 string.

jonashendrickx (approved):

src/AdminConsole/TagHelpers/LocalTime.cs
```diff
@@ -0,0 +1,16 @@
...
+[HtmlTargetElement("asp-local-time")]
+public class LocalTime : TagHelper
+{
+    [HtmlAttributeName("datetime")]
+    public DateTime DateTime { get; set; }
+
+    public override void Process(TagHelperContext context, TagHelperOutput output)
+    {
+        output.TagName = "local-time";
+        output.Attributes.Add("datetime", DateTime.SpecifyKind(DateTime, DateTimeKind.Utc).ToString("O"));
```
Tyrrrz, line 14: Should probably be `ToString("O", CultureInfo.InvariantCulture)`

jonashendrickx: If it's ISO 8601 it likely wouldn't matter.

PR #148, opened 2023-10-17
Add invariant culture to local time

jrmccannon:


jonashendrickx (commented):
It should always be iso 8601 compliant, even without it though.

PR #151, opened 2023-10-23
Added event logs for invalid and frozen keys.

jrmccannon:
There is a little bit of rework here for writing application events. This is due to the fact that we need to have an unauthenticated version of the event logger to write unauthenticated requests to the logs (invalid keys).

There is now a scoped `EventCache` object that holds all events that need to be logged to the database. There are two loggers now. The Unauthenticated logger will do a feature check before writing to the db. It extends the base one which does not have that check.


src/Api/Middleware/EventLogContextMiddleware.cs
```diff
@@ -15,19 +15,20 @@ public async Task InvokeAsync(HttpContext context, IEventLogContext eventLogCont
     {
         var tenantId = context.Request.GetTenantName();
         var requestKey = context.Request.GetPublicApiKey() ?? context.Request.GetApiSecret();
+        var authenticated = context.User.Identity?.IsAuthenticated ?? false;
```
Tyrrrz, line 18: Maybe `isAuthenticated` for consistency?

jonashendrickx (approved):

src/Api/Middleware/EventLogStorageCommitMiddleware.cs
```diff
@@ -5,18 +5,30 @@ namespace Passwordless.Api.Middleware;
...
 
-    public EventLogStorageCommitMiddleware(RequestDelegate next) => _next = next;
+    public EventLogStorageCommitMiddleware(RequestDelegate next, ILogger<EventLogStorageCommitMiddleware> logger)
+    {
+        _next = next;
+        _logger = logger;
+    }
 
-    public async Task InvokeAsync(HttpContext context, IEventLogger eventLogger)
+    public async Task InvokeAsync(HttpContext context, IServiceProvider provider)
```
Tyrrrz, line 16: Why resolve the logger like this? Is it because the service registration can change?

Tyrrrz (approved):

src/Service/EventLog/Loggers/EventCache.cs
```diff
@@ -0,0 +1,11 @@
...
+
+namespace Passwordless.Service.EventLog.Loggers;
+
+public class EventCache
+{
+    private readonly List<ApplicationEvent> _events = new();
+    public IEnumerable<ApplicationEvent> GetEvents() => _events;
+    public bool IsEmpty() => !_events.Any();
+    public void Add(ApplicationEvent applicationEvent) => _events.Add(applicationEvent);
+}
```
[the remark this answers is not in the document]
jrmccannon: Technically it would get emptied when it gets deleted by garbage collection when the request ends (since its scoped), but it should be explicit. Making the change now.

src/Api/Middleware/EventLogStorageCommitMiddleware.cs
jrmccannon: Since we're in middleware, the event logger that gets injected is outside the authentication context. this way, we can get the instance of the event logger that reflects the current state of the request.

abergs: That took a while to understand, nice catch Jared.

src/Service/Models/AccountMetaInformation.cs
```diff
@@ -24,4 +25,5 @@ public string AdminEmailsSerialized
     public DateTime? DeleteAt { get; set; }
 
     public AppFeature Features { get; set; }
+    public IEnumerable<ApplicationEvent> Events { get; set; }
```
abergs, line 28: @jrmccannon are we 100% that thie doesn't get loaded up when we load AccountMetaInformation from the db?

If they get loaded by default it will severely impact perfromance. Could we test?

src/Service/EventLog/Loggers/EventCache.cs
```diff
@@ -0,0 +1,12 @@
+using Passwordless.Service.EventLog.Models;
+
+namespace Passwordless.Service.EventLog.Loggers;
+
+public class EventCache
```
abergs, line 5: Is this cache used anywhere outside the EfWriter? If not - should it be a private subclass, not injected by DI? Seems to be to just be a implementation detail on how we do delayed writes to EF, no?

abergs (commented):
Well done Jared, I have some comments I'd like to adress before merging.

src/Service/Models/AccountMetaInformation.cs
Tyrrrz: Didn't notice this, but we should avoid using `IEnumerable<T>` properties because they might map to some lazily evaluated collection (e.g. from EF) and that might result in some deferred exceptions. A better option would be `IReadOnlyCollection<T>` or `IReadOnlyList<T>`.

abergs: @justindbaur Do you know if there is some way to "guarantee" we're not loading this?

src/Service/EventLog/Loggers/EventCache.cs
jrmccannon: By putting it in DI as scoped, it exists for the lifetime of the request. The request could pull different instances of the `IEventLogger` throughout the request depending on where we put logs.  This acts as a manager class I guess to hold the event collection for the lifetime of the request. The Unauthenticated Logger could be pulled, write an event for something, then move on. Later another thing could write an event and finish.  At the end of the request, we would ask for a final logger, and then commit everything from the Event Cache and clear it out. 

If we removed it from DI, then when the one instance was destroyed, then we'd lose the events held in that collection. Or am I missing something?

src/Api/Middleware/EventLogStorageCommitMiddleware.cs
jrmccannon: Added comment explaining why.

src/Service/Models/AccountMetaInformation.cs
justindbaur: `AccountMetadataInformation` has a one-to-many relationship with `ApplicationEvent`? Then you get the guarantee that EF won't load that by just NOT doing a `Include()` or `ThenInclude` on queries that would request it. Some people will make this property virtual to sort of signify that it's another table that isn't always there but that is purely optional. 

A lot of docs will make the property be an `ICollection<T>` https://learn.microsoft.com/en-us/ef/core/modeling/relationships/one-to-many but `IReadOnlyCollection` should be fine.

abergs: Thanks @justindbaur.
Let's add Virtual and we're good to go @jrmccannon

abergs (approved):

abergs:
@jrmccannon Just needs a format.

src/Service/Models/AccountMetaInformation.cs
jrmccannon: Done

src/Service/Storage/Ef/DbGlobalContext.cs
```diff
@@ -62,8 +62,14 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
-        modelBuilder.Entity<ApplicationEvent>()
-            .HasKey(x => x.Id);
+        modelBuilder.Entity<ApplicationEvent>(builder =>
+        {
+            builder.HasKey(x => x.Id);
+            builder.HasOne(x => x.Application)
+                .WithMany(x => x.Events)
+                .HasForeignKey(x => x.TenantId)
+                .IsRequired();
+        });
```
abergs, line 72: @jrmccannon This seems like we need migrations?

abergs (changes requested):
See comment on migrations

src/Service/Storage/Ef/DbGlobalContext.cs
jrmccannon: Done

src/Api/Extensions/HttpRequestExtensions.cs
```diff
@@ -5,41 +5,23 @@ namespace Passwordless.Api.Extensions;
...
-            return ApiKeyUtils.GetAppId(request.Headers["ApiSecret"].ToString());
-        }
+        var key = request.GetPublicApiKey() ?? request.GetApiSecret();
 
-        return null;
+        return key != null ? ApiKeyUtils.GetAppId(key) : null;
     }
-}
 
+    public static string? GetTenantName(this HttpRequest httpRequest) =>
```
jonashendrickx, line 19: Looks good.

abergs (approved):

PR #154, opened 2023-10-25
PAS203: Enterprise self-serve

jonashendrickx:
image


Tyrrrz:
FYI: fixed a merge conflict

src/AdminConsole/Migrations/Sqlite/20231024113214_BillingRefactoring.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        migrationBuilder.AddColumn<string>(
+            name: "BillingSubscriptionItemId",
+            table: "Organizations",
+            type: "TEXT",
+            nullable: true);
+
+        migrationBuilder.Sql("UPDATE Organizations SET BillingPlan = a.BillingPlan, BillingSubscriptionItemId = a.BillingSubscriptionItemId FROM Organizations o INNER JOIN Applications a ON o.Id = a.OrganizationId;");
+
+        migrationBuilder.DropColumn(
```
abergs, line 28: Breaking database change - not something we want to do. We're only "allowed" to add columns, since we must be able to apply migrations but then roll back to previous version of code, in case of bugs etc.

abergs (changes requested):

PR #162, opened 2023-10-30
CVE-2023-44270

jonashendrickx:


abergs (commented):
@jonashendrickx Some css-styles were removed, are you sure these are unused? 

Was this a result from running `npm run tw:build`?

jonashendrickx:
> @jonashendrickx Some css-styles were removed, are you sure these are unused?
> 
> Was this a result from running `npm run tw:build`?

I'm also getting the same changes otherwise forcing the same tailwind version. I might be the only one rebuilding the tailwind CSS. I verified the changes that nothing looked weird.

We could make it part of the build process and stop committing the generated stylesheets during BEEEP.

PR #166, opened 2023-10-30
Use GH Action v3 to deploy

abergs:


.github/workflows/main.yml
```diff
@@ -171,19 +171,26 @@ jobs:
         # Change directory into the app folder, so that the archive doesn't contain the folder itself
         working-directory: ${{ matrix.app }}/
         run: zip -r ../app.zip .
-
+      
+      # This step is ONLY needed when using the Azure CLI to either set app settings or publish
```
Tyrrrz, line 175: It's not needed for the next step?

PR #168, opened 2023-10-30
Try to configure DD Version from code

abergs:
We currently utilize DD for monitoring logs, traces, deployments etc.

A essential feature is the version-tagging offered by DataDog to discover regressions and perf differences between deployments. 

This PR tries to move the responsibility of providing the version value away from the environment variable `DD_VERSION` towards code configuration.

**Motivation:** Configuring the DD_VERSION environment variable makes deployment complex and brittle, and out of sync (Actual code version might not be same as ENV variable)

**Problem:** Configuring the Version this way does not work -- Data Dogs UI indicates we no longer provide a version together with the logs.

Note: We're using the Azure App Service DataDog Extension.
Note 2: We would like to move towards OpenTelemetry where possible.

### DIFF
--- a/src/AdminConsole/Program.cs
+++ b/src/AdminConsole/Program.cs
@@ -1,4 +1,3 @@
-using System.Reflection;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc.Infrastructure;
 using Microsoft.Extensions.Options;
--- a/src/AdminConsole/Program.cs
+++ b/src/AdminConsole/Program.cs
@@ -23,12 +23,6 @@ using Passwordless.Common.Services.Mail;
 using Serilog;
 using Serilog.Sinks.Datadog.Logs;
 
-// Set Datadog version tag through an environment variable, as it's the only way to set it apparently
-Environment.SetEnvironmentVariable(
-    "DD_VERSION",
-    Assembly.GetExecutingAssembly().GetName().Version?.ToString() ?? "unknown"
-);
-
 Log.Logger = new LoggerConfiguration()
     .WriteTo.Console()
     .CreateLogger();
@@ -77,9 +71,15 @@ void RunTheApp()
         IConfigurationSection ddConfig = ctx.Configuration.GetSection("Datadog");
         if (ddConfig.Exists())
         {
-            var version = Assembly.GetExecutingAssembly().GetName().Version?.ToString() ?? "unknown";
+            var version = assembly.GetInformationalVersion() ??
+                              assembly.GetName().Version?.ToString() ??
+                              "unknown version"; // 1.2.3-ci-SHA
+            
+            // TRY TO GET DataDog Azure App Service Extension to use a version configured from code
+            // instead of DD_VERSION environment variable
+            // Doesn't seem to work though? 
 
-            // setup tracing
+            // Configure the DataDog Tracer with version
             var settings = TracerSettings.FromDefaultSources();
             settings.ServiceVersion = version;
             Tracer.Configure(settings);
@@ -90,7 +90,6 @@ void RunTheApp()
                 config.WriteTo.DatadogLogs(
                     apiKey: ddKey,
                     tags: new[] { "version:" + version },
-                    service: "pass-admin-console",
                     configuration: new DatadogConfiguration(ddConfig.GetValue<string>("url")));
             }
         }
--- a/src/AdminConsole/Program.cs
+++ b/src/AdminConsole/Program.cs
@@ -1,6 +1,4 @@
 using System.Reflection;
-using Datadog.Trace;
-using Datadog.Trace.Configuration;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc.Infrastructure;
 using Microsoft.Extensions.Options;
@@ -18,7 +16,6 @@ using Passwordless.AdminConsole.Services.Mail;
 using Passwordless.AdminConsole.Services.PasswordlessManagement;
 using Passwordless.AspNetCore;
 using Passwordless.Common.Configuration;
-using Passwordless.Common.Extensions;
 using Passwordless.Common.Middleware.SelfHosting;
 using Passwordless.Common.Services.Mail;
 using Serilog;
@@ -69,30 +66,18 @@ void RunTheApp()
             config.WriteTo.Seq("http://localhost:5341");
         }
 
-        IConfigurationSection ddConfig = ctx.Configuration.GetSection("Datadog");
-        if (ddConfig.Exists())
+        var ddApiKey = Environment.GetEnvironmentVariable("DD_API_KEY");
+        if (!string.IsNullOrEmpty(ddApiKey))
         {
-            var assembly = Assembly.GetExecutingAssembly();
-            var version = assembly.GetInformationalVersion() ??
-                              assembly.GetName().Version?.ToString() ??
-                              "unknown version"; // 1.2.3-ci-SHA
+            var ddSite = Environment.GetEnvironmentVariable("DD_SITE") ?? "datadoghq.com";
+            var ddUrl = $"https://http-intake.logs.{ddSite}";
+            var ddConfig = new DatadogConfiguration(ddUrl);
 
-            // TRY TO GET DataDog Azure App Service Extension to use a version configured from code
-            // instead of DD_VERSION environment variable
-            // Doesn't seem to work though? 
-
-            // Configure the DataDog Tracer with version
-            var settings = TracerSettings.FromDefaultSources();
-            settings.ServiceVersion = version;
-            Tracer.Configure(settings);
-
-            var ddKey = ddConfig.GetValue<string>("ApiKey");
-            if (!string.IsNullOrWhiteSpace(ddKey))
+            if (!string.IsNullOrEmpty(ddApiKey))
             {
                 config.WriteTo.DatadogLogs(
-                    apiKey: ddKey,
-                    tags: new[] { "version:" + version },
-                    configuration: new DatadogConfiguration(ddConfig.GetValue<string>("url")));
+                    ddApiKey,
+                    configuration: ddConfig);
             }
         }
     });
--- a/src/AdminConsole/Program.cs
+++ b/src/AdminConsole/Program.cs
@@ -1,3 +1,4 @@
+using System.Reflection;
 using Datadog.Trace;
 using Datadog.Trace.Configuration;
 using Microsoft.AspNetCore.Authorization;
@@ -17,6 +18,7 @@ using Passwordless.AdminConsole.Services.Mail;
 using Passwordless.AdminConsole.Services.PasswordlessManagement;
 using Passwordless.AspNetCore;
 using Passwordless.Common.Configuration;
+using Passwordless.Common.Extensions;
 using Passwordless.Common.Middleware.SelfHosting;
 using Passwordless.Common.Services.Mail;
 using Serilog;
@@ -70,6 +72,7 @@ void RunTheApp()
         IConfigurationSection ddConfig = ctx.Configuration.GetSection("Datadog");
         if (ddConfig.Exists())
         {
+            var assembly = Assembly.GetExecutingAssembly();
             var version = assembly.GetInformationalVersion() ??
                               assembly.GetName().Version?.ToString() ??
                               "unknown version"; // 1.2.3-ci-SHA
--- a/src/AdminConsole/Program.cs
+++ b/src/AdminConsole/Program.cs
@@ -1,3 +1,6 @@
+using System.Reflection;
+using Datadog.Trace;
+using Datadog.Trace.Configuration;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc.Infrastructure;
 using Microsoft.Extensions.Options;
@@ -19,6 +22,12 @@ using Passwordless.Common.Middleware.SelfHosting;
 using Passwordless.Common.Services.Mail;
 using Serilog;
 using Serilog.Sinks.Datadog.Logs;
+
+// Set Datadog version tag through an environment variable, as it's the only way to set it apparently
+Environment.SetEnvironmentVariable(
+    "DD_VERSION",
+    Assembly.GetExecutingAssembly().GetName().Version?.ToString() ?? "unknown"
+);
 
 Log.Logger = new LoggerConfiguration()
     .WriteTo.Console()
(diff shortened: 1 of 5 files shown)


abergs:
I've tried bringing this up with the Datadog Support but so far without luck.

I realize this is not the correct or official channel, but I was wondering if perhaps you have any pointers @andrewlock?

This is our third attempt at this... We're at a loss and scratching our heads if this is even possible?

src/AdminConsole/Program.cs
```diff
@@ -68,11 +72,26 @@ void RunTheApp()
...
+            settings.ServiceVersion = version;
+            Tracer.Configure(settings);
+
             var ddKey = ddConfig.GetValue<string>("ApiKey");
             if (!string.IsNullOrWhiteSpace(ddKey))
             {
                 config.WriteTo.DatadogLogs(
                     apiKey: ddKey,
+                    tags: new[] { "version:" + version },
                     configuration: new DatadogConfiguration(ddConfig.GetValue<string>("url")));
```
Tyrrrz, line 95: It would be ideal if this were automatic, but the fact that even just configuring the version manually doesn't work is really weird 🤔

andrewlock:
Hey @abergs - sorry to see you're having so many issues 🙁 It looks like you're hitting a bunch of rough edges, so I'll try to explain _why_ you're hitting them. (Sorry for the length, this is part opinion piece, part background, part support!)

I'll start of by saying that I _always_ suggest using environment variables where possible. There are a whole bunch of technical reasons for it (plus my personal opinions, 12 factor apps and all that), but the biggest ones are

1. They always work (we have to do some config on the native side, and there the _only_ config we support is env variables). 
2. They mean we don't have to do any expensive re-initialization. We have to hook in very early in your app's lifecycle, and we initialize everything. If you then call `Tracer.Configure()` we have to re-do all of that.
3. The fact we have to do re-initialization with config in code means that it's possible for some long-lived singleton components to use the "original" config, and not the rebuilt config. This leads to inconsistencies, similar to what you're describing. 

That all said, I do understand they can be problematic in some cases, and things should still _work_ 🤦‍♂️

Anyway, that's the background, I'll try to walk through each of your points.

> Configuring the DD_VERSION environment variable makes deployment complex and brittle, and out of sync (Actual code version might not be same as ENV variable)

Fundamentally, this _shouldn't_ happen in AAS as I understand it 🤔 If you're explicitly setting an application setting on the webapp using `az webapp`, I don't understand _why_ that's not being applied 🤷‍♂️  What's more you're doing a push subsequently, so the app presumably has to be reloaded, so it shouldn't be like an old process hanging on to an old value 🤔  I'm not overly familiar with AAS tbh, but this feels odd. Could you expand on how/why that's not working? It's keeping the old value?

>  `// TRY TO GET DataDog Azure App Service Extension to use a version configured from code`
>  `// instead of DD_VERSION environment variable`

This maybe hits those rough edges I talked about. Currently there are some settings that we read from the initial settings and _don't_ allow them to be "changed" in code in general, and part of that relates to log injection, because of the expensive one-time setup we do.

However, as far as I can tell, [we always fetch the _latest_ setting versions (allowing for re-initialization)](https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/Logging/Serilog/LogsInjection/LoggerDispatchInstrumentation.cs#L49-L56), when reading the tracer version. So it _should_ be injecting the `version` value specified in `settings.ServiceVersion = version`. I would need to check the debug logs to see exactly what's going on there. Is it possible `version` is null here? That's the only reason I can think of for why this wouldn't be working 🤔 

> but the fact that even just configuring the version manually doesn't work is really weird

There's actually another complicating setup in your case, because [the Datadog Serilog sink](https://github.com/DataDog/serilog-sinks-datadog-logs) has completely separate setup. It's managed by a different team which makes it a bit trickier for me to suggest the solution, but looking through the code, there doesn't seem to be a way to explicitly set the `version` tag at all 🙁 [I've raised an issue about that here](https://github.com/DataDog/serilog-sinks-datadog-logs/issues/108). The `tags: new[] { "version:" + version },` approach won't work as best as I can tell (explained in that issue). 

(Previously it didn't _matter_ that there's no way to set the version, because we explicitly inject that into the logs using the .NET tracer, but if that didn't work (e.g. because `version` is null) then this wouldn't either. Although... if `version` is null it's irrelevent I guess😅)

As an aside, [you _could_ consider agentless logging](https://docs.datadoghq.com/logs/log_collection/csharp/?tab=serilog#enable-agentless-logging) instead of using the Datadog Serilog sink. With this solution the .NET tracer takes care of adding the sink, and adding all the tags etc. We then automatically hook up a sink for you, without needing the extra Datadog sink packages. As you're using AAS (and so have already set `DD_API_KEY` and `DD_SITE`, you would just need to [set this one env var to enable it](https://docs.datadoghq.com/logs/log_collection/csharp/?tab=serilog#enable-agentless-logging):

```
DD_LOGS_DIRECT_SUBMISSION_INTEGRATIONS=Serilog
```

And then you could remove the extra sink, and rely on the central tracer configuration. Of course, that still requires that `DD_VERSION` is set correctly, which appears to not be the case 😞 

> It would be ideal if this were automatic

In terms of automatically pulling the `DD_VERSION` tag from assembly attributes, that's a feature we've discussed, and _in theory_ we're on board. The problem is that we can't guarantee _when_ we can actually do it. Because of the way we inject into the process, we may initialize _before_ your DLL is loaded, so we won't necessarily find the config. It also means we have to keep checking to see if we can find it, and have to "guess" what the entry assembly is (which isn't necessarily always obvious, e.g. with aspnet/iis etc). All of which makes it less attractive than being able to set it somewhere statically.

**Alternative approach**

So, finally, I have a concrete alternative you could consider. By default, [we read configuration from multiple sources](https://docs.datadoghq.com/tracing/trace_collection/library_config/dotnet-core/?tab=environmentvariables): code, env variables, web.config/app.config (for .NET FX) and datadog.json.  All but "code" are static, so easy for us to ensure our features all stay in sync.

If it's easier for you, you could consider deploying a _datadog.json_ file with your app:

```json
{ "DD_VERSION": "1.2.3" }
```

You can likely output this file pretty easily as part of your build/deploy process. You then wouldn't need to do the configuration in code, plus avoid any AAS weirdness around settings. Maybe the best of both worlds?

All that's a bit unsatisfying, especially as I don't feel like the root cause in either case is obvious so I'll try to escalate it on our side. The most useful thing would be to enable `DD_TRACE_DEBUG=1` on an app that displays the behaviour (and restart it) so that we can grab all the debug logs to figure out what's going on. But if the datadog.json solution works for you, I also understand if you don't want the hassle 😄 

Again, sorry for the verbosity, but hope this helps 🤞

Tyrrrz:
@andrewlock Hi, thank you for such a detailed reply. @abergs knows more about DD integration so he'll probably comment on other points. However, I wanted to explain why we prefer to avoid using an environment variable to set the version:

1. The version is set differently depending on whether it's a regular CI build or a release build. Of course, it's still possible to determine which version is getting passed, but it's not trivial and requires extra steps, output routing, and adds measurable complication to the workflow. On the other hand, the assembly itself knows what version it is, so it makes a lot of sense to try and set it from code -- this is (almost) as trivial as `Assembly.GetExecutingAssembly().GetName().Version`.
2. There appears to be no way (from what we tested) to set an environment variable in AAS for just one specific deployment. The env is set for the application as a whole and it stays there until it's removed or replaced. That means that we can get out-of-sync issues where the variable is set to a value that doesn't correspond to reality, if something goes subtly wrong in the deployment.
3. The whole reason we're even doing this in the first place, is because DataDog doesn't seem to be able to infer the version of the assembly automatically for some reason. In an ideal world, we shouldn't have to set the version neither through variables or through code, it should be populated by the integration, right?

andrewlock:
> The env is set for the application as a whole 

Ah ok, the deployment slots issue makes sense, that's something I've heard about before, just didn't connect. You can create slot-specific settings though right? I understand if it makes things more complex, just want to make sure we're not missing an easy-win. 

> On the other hand, the assembly itself knows what version it is, so it makes a lot of sense to try and set it from code

Yeah, and that _should_ work 🤔 As much as I dislike configuration in code for DD for the various reasons I laid out, it should be working for you AFAICT. I don't understand why it's not, I think we'd have to get debug logs through the escalation to figure that one out.

> The whole reason we're even doing this in the first place, is because DataDog doesn't seem to be able to infer the version of the assembly automatically for some reason. In an ideal world, we shouldn't have to set the version neither through variables or through code, it should be populated by the integration, right?

This is the point I was trying to address in my previous post. I agree, this would be nice, and ideally would be the default behaviour, but there are some complexities. The biggest problem is that we (the .NET tracer) configure ourselves potentially _before_ the entry assembly has loaded. That would mean the correct version is not available until later, which 1. means we need to keep checking if it's available, which adds overhead etc. and 2. _potentially_ causes versions to drift out of sync. We're not ruling it out, it's just it's not such an obvious easy-win as it initially seems.

abergs:
@andrewlock Thank you for your incredible answer. It helps me put things into perspective and also confirm my suspicion that we found ourself in a intersection of rough edges.

While using the assembly version would have been elegant, I think we might go with the datadog.json file -- it effectively solves our problem of drift and the other, main offender: Complicated deployment and being a typo from logging all Azure App Setting values.

In addition to being "drifty" - having env variables in azures leads to more problems: It requires additional steps (than could malfunction) in our workflow as well as giving the CI/CD pipeline more access to our cloud environment than I'm comfortable with.
**Especially problematic** since how the `az webapp config appsettings set` CLI API is designed to return/print all app settings, regardless if you're just setting one, which discloses information about our deployment environment into the GH Action Logs. Currently fixed by a ` > /dev/null` - but to me it both feels and looks like a a high resolution footgun.


Aside: If I understand you correctly, we could remove all `DataDog`-packages and rely solely on the combination of environment variables (`DD_ENV, DD_SITE`) + `datadog.json` (`DD_VERSION`).

Any other pointers for having a great OpenTelemtry experience? I know we're pushing hard in that direction across Bitwarden, so we're happy to learn from perhaps one of the best when it comes to .NET Tracing :)

Again, thank you kindly for your assistance.

Tyrrrz:
Thank you again @andrewlock. Just to clarify, the `version` passed down is not null (that's taken care of by the `??` operator) and not empty.

abergs:
@andrewlock Again, very appreciated.

We currently use custom tags on traces, but we also set them on Activity: https://github.com/passwordless/passwordless-server/blob/997bf989f0f76d72724948c990be8c6b1a87ff69/src/Api/Helpers/LoggingMiddleware.cs#L37C26-L37C26
Moving towards OTEL will as far as I understand, only remove one line. But that's fine -- it's more standardised.

Do tags/baggage on `Acitivity` carry over to the spans recorded into DD? (aside: I wonder if I really need both tag+baggage)

I will update this PR with the new approach, beeing:

* Use the Azure Extension
* Some things in environ
* Write version to datadog.json during CI/CD
* If `Activity` tags are supported in DD, Enable OTEL, `DD_TRACE_OTEL_ENABLED=1` and remove the Trace Package
* Keep the DataDog sink, to be able to filter on namespaces/levels. If we at some point don't want those knobs, we could enable agentless with

andrewlock:
> Do tags/baggage on Acitivity carry over to the spans recorded into DD? (aside: I wonder if I really need both tag+baggage)

Tags added to `Activity` do carry over to the spans, yes, but we don't currently support baggage.

RE moving to OTel, one thing to be aware of is that your traces may well change depending on the other libraries you have and _their_ support for `Activity` etc. For example, if we automatically instrument a library and they later add direct OTel support, then you may get "double spans" with `DD_TRACE_OTEL_ENABLED`. We're actively trying to address those issues if they appear though, so do let us know if you run into any issues, as we'd love to address any rough edges you have with the Otel support!

abergs:
@andrewlock Thought I'd update you with a progress report, as we have some suprises.

We decided to use `datadog.json` and ENV variables the following way:

We store this in our src, as default, and update DD_VERSION during CI.
```json
{
    "DD_SITE": "datadoghq.eu",
    "DD_ENV": "pass-any",
    "DD_VERSION": "0.0.0",
    "DD_LOGS_INJECTION": true
}
``` 

In Azure App Settings, we store:

```env
DD_API_KEY=abc
DD_ENV=pass-devtest
```

We've kept both of the nuget packages (Tracing + Serilog) for now.
I've updated the .NET Datadog APM Azure Extension to `2.40.0`.


Result:
* Version tagging worked great for logs! (We need to manually provide configuration to the Sink though)
* Tracing stopped working.

Turns out that the `DD_SITE` from datadog.json was not respected/read by the Azure Extension(?).
Adding it as as environment variable worked - and we now get tracing sent to DD.

Very surprising. It still feels like something is a bit off with our setup... not sure yet.

andrewlock:
Hey @abergs, glad you're making progress at least!

> Turns out that the DD_SITE from datadog.json was not respected/read by the Azure Extension(?).

Yes, unfortunately, the agent (running in the Azure Extension) is "separate" in a lot of ways, and primarily needs configuring with environment variables... It makes me wonder if we can do something about that though, I'll discuss with the team 🤔

> Tracing stopped working.

Just to confirm, that was fixed by adding `DD_SITE` to the env vars right?

> We need to [manually provide configuration](https://github.com/passwordless/passwordless-server/blob/main/src/Api/Program.cs#L47) to the Sink though

Yeah, this is where the automatic agentless instrumentation may work well for you, seeing as you don't have any custom config AFAICT, but then again, if it's working for you... 😅 

Thinking about the issue further, I'm fairly sure your problems arrived because you were in a "version conflict" scenario where the Datadog.Trace NuGet package and the Azure Extension were at different versions. When that happens, we have to do a lot of work to try to "unify" the dlls which adds overhead and causes potential weirdness like the issues you _originally_ saw with configuration in code. 

So while having different versions _works_, we _strongly_ recommend keeping the version numbers in sync. We're well aware this can be operationally difficult though, and we have some bigger changes in the planning phase to fix this mismatch issue once and for all. 

Don't hesitate to reach out if you run into any more issues!

abergs:
@andrewlock I'm thankful and impressed by your engagement in this. Thank you for caring.

> Just to confirm, that was fixed by adding DD_SITE to the env vars right?

Yes. 

> Yeah, this is where the automatic agentless instrumentation may work well for you, seeing as you don't have any custom config AFAICT, but then again, if it's working for you... 😅

Those exact thoughts went through my head. I just kept to the "it's working, don't touch it" route.

> Thinking about the issue further, I'm fairly sure your problems arrived because you were in a "version conflict" scenario

I see! I didn't understand that was a strict requirement. Will try to keep them synced.

abergs:
@andrewlock Just a thought: If our hypothesis is that the Azure Extension is somewhat one of the main characters in this saga, is there a better way for us to run APM?

We've been looking to run Azure App Services based on our docker container, but I couldn't make that work with DD. My idea was to put a DD Agent inside the app container, but couldn't figure out if that was good/possible.

andrewlock:
Glad it's working @abergs! 😄 

> I see! I didn't understand that was a strict requirement. Will try to keep them synced.

Yeah, it's one of those things that [we call out in the docs](https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/dotnet-framework/?tab=windows#custom-instrumentation), but it's sometimes easier said than done. As I say, we're working towards a "once and for all" fix for the issue soon that should make things easier. Until then, you definitely want to keep them in sync!

> If our hypothesis is that the Azure Extension is somewhat one of the main characters in this saga, is there a better way for us to run APM?

If you're running in Azure App Services, then you'll need to use the extension, one way or another, as it takes care of some of the oddities of AAS for a start. Also, you really don't want to run the agent in the same container as your app AFaIK (if that's even possible - I haven't tried it tbh!).

Just to be clear, we support both AAS on Windows [and Linux now](https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_linux?tab=nodenetphppython), though I'm not very familiar with the Linux setup, so will defer to the docs on that one.

Glad you've got it working anyway, as I say, any further issues just let us know 🙂

PR #175, opened 2023-11-01
Add docker image for the API project

Tyrrrz:


Api.dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN dotnet publish src/Api/ \
+    --configuration Release \
+    --self-contained \
+    --use-current-runtime \
+    --output src/Api/bin/publish/
+
+# ** Run
+
+# Use `runtime-deps` instead of `runtime` because we have a self-contained assembly
+FROM mcr.microsoft.com/dotnet/runtime-deps:7.0-alpine AS run
```
Tyrrrz, line 20: Using a self-contained build and alpine for the smallest run-time image possible (i.e. faster test startup if pull is required)

Api.dockerfile
```diff
@@ -0,0 +1,32 @@
...
+WORKDIR /opt/pwdls/
+
+EXPOSE 80
+EXPOSE 443
+
+# Alpine image doesn't come with the ICU libraries pre-installed, so we need to install them manually.
+# Technically, we shouldn't need globalization support in the API, but some EF queries fail without it at the moment.
+RUN apk add --no-cache icu-libs
+ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=false
```
Tyrrrz, line 29: This is something we may consider looking into in the future. It would make the image even smaller and will probably speed up the app (in production too) as well.

abergs:
@Tyrrrz Nicely done, should we mention in the Dockerfile that this image is only intended for testing? Since we will be using the "unified" image for actually self-hosting.

For deployment strategy, we could start by putting on Github Registry, but we should later push it to dockerhub. I do however think we should name it so it is related to testing, not to confused self-hosters.

abergs (approved):

Tyrrrz:
@abergs

> should we mention in the Dockerfile that this image is only intended for testing

Do you mean as a `LABEL` or something else?

> For deployment strategy, we could start by putting on Github Registry, but we should later push it to dockerhub. I do however think we should name it so it is related to testing, not to confused self-hosters.

Can do. However, I'm not sure if the GitHub's Docker Registry works the same as NuGet Registry, where it requires authentication in order to pull packages – in which case deploying straight to DockerHub might be easier for both us and the potential users long term.

As for the name, do you have any ideas? `Passwordless.Server.Api.TestServer`?

abergs:
@Tyrrrz Either a LABEL or just a comment section in the top.

Agree that DockerHub is the long term solution, but building nightly ones to Github is also a nice start.

Final image could potentially be `bitwarden/passwordless-test-server`?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Build & push image
+        run: >
+          docker buildx build
+          --file ${{ matrix.app }}.dockerfile
+          --platform linux/amd64,linux/arm64
+          --push
+          --tag passwordless/api-test-server:latest
+          ${{ github.event_name == 'release' && '--tag passwordless/api-test-server:$GITHUB_REF_NAME' || '' }}
+          ${{ github.event_name == 'release' && '--tag passwordless/api-test-server:stable' || '' }}
+          .
```
Tyrrrz, line 45: Using `buildx` here to build for multiple platforms at the same time and to pack them all in one image bundle.

Tyrrrz: The tags are set accordingly:

`:latest` -- last `main` branch commit or release
`:stable` -- last release
`:1.2.3` -- specific release

Api.dockerfile
```diff
@@ -0,0 +1,40 @@
+LABEL description="Docker image of the Passwordless API, intended solely for development and integration testing purposes."
```
Tyrrrz, line 1: Label to signify that this is not intended for general use

.github/workflows/docker.yml
```diff
@@ -0,0 +1,45 @@
...
+          echo ${{ secrets.GITHUB_TOKEN }} |
+          docker login ghcr.io --username ${{ github.actor }} --password-stdin
+
+      - name: Build & push image
+        run: >
+          docker buildx build
+          --file ${{ matrix.app }}.dockerfile
+          --platform linux/amd64,linux/arm64
+          --push
+          --tag ghcr.io/passwordless/api-test-server:latest
```
Tyrrrz, line 42: This or `bitwarden/passwordless-api-test-server`, @abergs? It needs to highlight that it's `api`

abergs: Let's not use /bitwarden until it's ready. Passwordless/passwordless-test-api ? (And then we will just switch to bitwarden/ later)

PR #179, opened 2023-11-07
Integration tests with Test containers

jrmccannon:
This adds integration tests that use test containers to spin up MS SQL database.  It also includes testing the user registration flow.


src/Api/Api.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <UserSecretsId>passwordless-api</UserSecretsId>
+      <DockerDefaultTargetOS>Linux</DockerDefaultTargetOS>
```
Tyrrrz, line 5: What is this used for?

src/Api/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+WORKDIR "/src/src/Api"
+RUN dotnet build "Api.csproj" -c Release -o /app/build
+
+FROM build AS publish
+RUN dotnet publish "Api.csproj" -c Release -o /app/publish /p:UseAppHost=false
+
+FROM base AS final
+WORKDIR /app
+COPY --from=publish /app/publish .
+ENTRYPOINT ["dotnet", "Passwordless.Api.dll"]
```
Tyrrrz, line 22: Could probably use the `Api.dockerfile` in root, right?
Also, form what I understand, you don't actually use this image?

tests/Api.Integration.Tests/RegisterTests.cs
```diff
@@ -0,0 +1,203 @@
...
+        .RuleFor(x => x.TokenId, Guid.Empty);
+
+    public RegisterTests(PasswordlessApiFactory apiFactory)
+    {
+        _client = apiFactory.CreateClient();
+        _client.DefaultRequestHeaders.Add("User-Agent", "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/118.0.0.0 Safari/537.36");
+    }
+
+    [Fact]
+    public async Task Server_Can_Successfully_Retrieve_Token_For_Registration()
```
Tyrrrz, line 38: Need to include this `xunit.runner.json`: https://github.com/passwordless/passwordless-dotnet/blob/24661e9bbc064b6ecfae8de7a31393655517d170/tests/Passwordless.Tests/xunit.runner.json#L3-L4

And also no need to capitalize every word. Also consider this naming: `I_can_retrieve_registration_token`.

Here's how an analogous test looks in `Passwordless.NET`: https://github.com/passwordless/passwordless-dotnet/blob/24661e9bbc064b6ecfae8de7a31393655517d170/tests/Passwordless.Tests/TokenTests.cs#L16-L30

tests/Api.Integration.Tests/RegisterTests.cs
```diff
@@ -0,0 +1,203 @@
...
+    {
+        var request = TokenGenerator.Generate();
```
Tyrrrz, line 40: Consider splitting the test with `// Arrange`/`// Act`/`// Assert` for better visual clarity

tests/Api.Integration.Tests/RegisterTests.cs
```diff
@@ -0,0 +1,203 @@
...
+        var registerTokenResponse = await response.Content.ReadFromJsonAsync<RegisterEndpoints.RegisterTokenResponse>();
+        registerTokenResponse.Should().NotBeNull();
+        registerTokenResponse!.Token.Should().StartWith("register_");
+    }
+
+    [Fact]
+    public async Task Client_Can_Begin_Registration_Of_User_Credential_With_Server()
+    {
+        var tokenRequest = TokenGenerator.Generate();
+        var tokenResponse = await _client.AddSecretKey().PostAsJsonAsync("/register/token", tokenRequest);
```
Tyrrrz, line 54: Maybe the client should be generated per-test (instead of shared in the class), so that way you don't accidentally retain the old `SecretKey`/`ManagementKey` value?

Kinda similar to this: https://github.com/passwordless/passwordless-dotnet/blob/24661e9bbc064b6ecfae8de7a31393655517d170/tests/Passwordless.Tests/TokenTests.cs#L20

tests/Api.Integration.Tests/RegisterTests.cs
```diff
@@ -0,0 +1,203 @@
...
+            .SetProtocol(VirtualAuthenticatorOptions.Protocol.CTAP2)
+            .SetHasResidentKey(true);
+
+        var options = new ChromeOptions();
+        options.AddArguments("--no-sandbox", "--disable-dev-shm-usage", "--headless");
+        var driver = new ChromeDriver(options);
+        driver.Url = originUrl;
+        driver.AddVirtualAuthenticator(virtualAuth);
+        var result = driver.ExecuteScript(GetScript(sessionResponse!.Data.ToJson()));
+
```
Tyrrrz, line 109: This could probably be abstracted away into something like `VirtualBrowser.PerformAuth(string sessionJson)`

tests/Api.Integration.Tests/RegisterTests.cs
```diff
@@ -0,0 +1,203 @@
...
+            Response = parsedResult
+        });
+
+        registerCompleteResponse.StatusCode.Should().Be(HttpStatusCode.OK);
+        var registerCompleteToken = await registerCompleteResponse.Content.ReadFromJsonAsync<TokenResponse>();
+        registerCompleteToken!.Token.Should().StartWith("verify_");
+    }
+
+    private static string GetScript(string jsonResponse) => $$"""
+       let registration = {{jsonResponse}};
```
Tyrrrz, line 128: Can use language injections here (`lang=javascript`) to get syntax highlighting in Rider. Might not work with interpolated strings though.

```suggestion
    private static string GetScript(string jsonResponse) =>
       // lang=javascript
       $$"""
       let registration = {{jsonResponse}};
```

Stretch goal: maybe refactor this into an embedded resource `.js` file, load it from memory, and provide all inputs through function parameters.

src/Api/Api.csproj
jrmccannon: Ah forgot about that. Good catch. This was auto-generated when I added docker support for API. I can roll those back.

Tyrrrz:
Also, we now have `Api.IntegrationTests` and `Api.Integration.Tests`. Probably would be nice to conslidate them, or replace the old tests with these new tests entirely.

tests/Api.Integration.Tests/RegisterTests.cs
```diff
@@ -0,0 +1,203 @@
...
+        .RuleFor(x => x.UserVerification, "Preferred")
+        .RuleFor(x => x.Aliases, x => new HashSet<string> { x.Person.FirstName })
+        .RuleFor(x => x.AliasHashing, false)
+        .RuleFor(x => x.ExpiresAt, DateTime.UtcNow.AddDays(1))
+        .RuleFor(x => x.TokenId, Guid.Empty);
+
+    public RegisterTests(PasswordlessApiFactory apiFactory)
+    {
+        _client = apiFactory.CreateClient();
+        _client.DefaultRequestHeaders.Add("User-Agent", "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/118.0.0.0 Safari/537.36");
```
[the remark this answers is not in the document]
jrmccannon: Correct. Wanted to leave a lot of this raw for the demo and would clean it up later.

Tyrrrz:
Where do Api logs currently go in tests? Would probably be nice to have them included in the test output, so that it's easier to debug issues on CI.

Might be helpful: https://github.com/martincostello/xunit-logging

tests/Api.Integration.Tests/RegisterTests.cs
jrmccannon: If I remember correctly, the constructor of the class will run for each fact. So each of them is getting a new client and isn't shared across the different facts.

tests/Api.Integration.Tests/RegisterTests.cs
jrmccannon: Leaving it raw for the demo and will clean it up later.

jonashendrickx (changes requested):

tests/Api.Integration.Tests/AuthorizationTests.cs
```diff
@@ -2,10 +2,12 @@
...
 using Microsoft.AspNetCore.Routing.Patterns;
 using Microsoft.Extensions.DependencyInjection;
+using Passwordless.Api.Integration.Tests.Helpers;
+using Xunit;
 
-namespace Passwordless.Api.IntegrationTests;
+namespace Passwordless.Api.Integration.Tests;
 
-public class AuthorizationTests : IClassFixture<TestWebApplicationFactory<Program>>
+public class AuthorizationTests : IClassFixture<PasswordlessApiFactory>
```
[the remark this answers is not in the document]
jrmccannon: Leaving as is. Will rewrite later.

tests/Api.Integration.Tests/Endpoints/App/AppTests.cs
```diff
@@ -0,0 +1,200 @@
...
+using Passwordless.Service.Models;
+using Passwordless.Service.Storage.Ef;
+using Xunit;
+
+namespace Passwordless.Api.Integration.Tests.Endpoints.App;
+
+public class AppTests : IClassFixture<PasswordlessApiFactory>, IDisposable
+{
+    private readonly HttpClient _client;
+    private readonly PasswordlessApiFactory _factory;
```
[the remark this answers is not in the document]
jrmccannon: I tried moving the dispose method and parts out to a base class. However, this gets messy because each class configures its own client separately and factory isn't required on every test so I think I'll just keep them separate for now and maybe implement a common class later.

tests/Api.Integration.Tests/Helpers/HttpClientTestExtensions.cs
```diff
@@ -0,0 +1,39 @@
+namespace Passwordless.Api.Integration.Tests.Helpers;
+
+public static class HttpClientTestExtensions
+{
+    public const string ApiKey = "test:public:2e728aa5986f4ba8b073a5b28a939795";
+    public const string ApiSecret = "test:secret:a679563b331846c79c20b114a4f56d02";
+
```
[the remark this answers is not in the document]
jrmccannon: These are the default keys from the Api used when seeding the database for local development.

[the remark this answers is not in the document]
Tyrrrz: Is there merit to apply a similar approach as done here, so as to not rely on defaults in case they change?

https://github.com/bitwarden/passwordless-dotnet/blob/f37c27bfc50c65bc409aa5507fa1907a6fb9d269/tests/Passwordless.Tests.Infra/TestApi.cs#L85-L125

Or not worth the effort?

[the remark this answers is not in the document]
jrmccannon: I think its a good next iteration. Right now, its a little tied to the database seed script. But, we can add it later. I started down the path, and more and more things started to be added to make it work seamlessly. Its a good add, but let's get this in place.

abergs:
Pinging @jonashendrickx for approval

jonashendrickx (approved):

tests/Api.IntegrationTests/Api.IntegrationTests.csproj
```diff
@@ -1,21 +1,41 @@
...
+  <ItemGroup>
+    <None Update="Helpers\Js\converters.js">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Update="Helpers\Js\createCredential.js">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Update="Helpers\Js\getCredential.js">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
```
Tyrrrz, line 38:
```suggestion
    <None Update="Helpers\Js\converters.js" CopyToOutputDirectory="PreserveNewest" />
    <None Update="Helpers\Js\createCredential.js" CopyToOutputDirectory="PreserveNewest" />
    <None Update="Helpers\Js\getCredential.js" CopyToOutputDirectory="PreserveNewest" />
```

PR #182, opened 2023-11-09
PAS195: Improve flow cancel pending deletion by e-mail

jonashendrickx:
E-mails are now being sent out by the administration console instead of the API, following the design elsewhere.

Magic link generation has been refactored out of the `MailService` class, so we can build magic links for different purposes such as sending out e-mails for application deletion cancellation magic links.

Clicking the magic link takes the user to the Organization/Overview. To have it go directly to the application settings page, additional work is needed given the way we handle route handling currently.

Endpoint on the Api is moved and now secured by the management key as well.

Summary of things to address:

- AdminEmails should be stored in AdminConsole
- Take a look at route handler for applications in admin console so we can make it work for magic links.


src/AdminConsole/Services/MagicLinks/MagicLinkSignInManager.cs
```diff
@@ -1,73 +1,49 @@
...
         : base(userManager, contextAccessor, claimsFactory, optionsAccessor, logger, schemes, confirmation)
     {
+        _magicLinkBuilder = magicLinkBuilder;
         _mailService = mailService;
-        _urlHelperFactory = urlHelperFactory;
-        _actionContextAccessor = actionContextAccessor;
         _eventLogger = eventLogger;
     }
 
     public async Task<SignInResult> SendEmailForSignInAsync(string email, string? returnUrl)
```
abergs, line 35: This result is never used, I think we can remove it since it doesn't add any value and is always successful.

src/AdminConsole/Services/Mail/DefaultMailService.cs
```diff
@@ -98,4 +94,65 @@ public Task SendOrganizationDeletedAsync(string organizationName, IEnumerable<st
...
+
+        await _provider.SendAsync(message);
+    }
+
+    public async Task SendApplicationToBeDeletedAsync(Application application, string deletedBy, string cancellationLink, ICollection<string> emails)
+    {
+        MailMessage message = new()
+        {
+            To = emails,
+            Bcc = new List<string> { "account-deletion@passwordless.dev" },
```
abergs, line 134: We can remove this BCC while we are at it.

abergs (approved):
Except for my small comment, it all looks good!

src/AdminConsole/Services/MagicLinks/MagicLinkSignInManager.cs
jonashendrickx: I'll add an item to investigate this

PR #183, opened 2023-11-09
PAS209: More padding needed around pagination breadcrumbs

jonashendrickx:
image

image


abergs:
Should we perhaps remove the underline styling on those numbers?

abergs (approved):

jonashendrickx:
> Should we perhaps remove the underline styling on those numbers?

They use a global styling class 'link-blue'. I've removed it from that class.

PR #185, opened 2023-11-09
PAS210: Creation of an admin passkey no longer triggers success message

jonashendrickx:
Given we're using a partial view to add credentials, we're adding the query parameters to the current url and then refresh the current page in Javascript. We're then parsing again the query parameters in the partial view. Allowing us to embed it easily on any page we please.

This pull request fixes 3 other open bugs.


src/AdminConsole/Pages/Shared/_AddPasskeys.cshtml
```diff
@@ -18,9 +24,9 @@
         </fieldset>
     </form>
 
-    <div class="status mt-3" v-cloak>
-        <SuccessAlertBox v-if="feedback.status =='completed'" message="{{feedback.message}}"></SuccessAlertBox>
-        <DangerAlertBox v-if="feedback.status == 'error'" message="{{feedback.message}}"></DangerAlertBox>
+    <div asp-if="hasCredentialAddFeedback" class="status mt-3" v-cloak>
+        <success-alert-box asp-if="@(credentialAddStatus == "completed")" message="@credentialAddMessage"></success-alert-box>
+        <danger-alert-box asp-if="@(credentialAddStatus == "error")" message="@credentialAddMessage"></danger-alert-box>
```
abergs, line 29: Hmm, is there another way to set the message that does not involve reading data provided in the URL? It's a foot gun when it comes to introducing XSS vulnerabilities and would be easier to just avoid it rather than doing the right encoding (and maintaining that encoding).

jonashendrickx: @abergs It would involve redesigning how it works completely. Because you also need to refresh the list of credentials. I did now add HtmlSanitizer.

abergs:
@jonashendrickx I was going to say that I'd really like for us to avoid using URL query input. Which I still my opinion, but seems like we don't need to sanitize, Razor already does it for us as long as we render using the `@` directive. 

https://learn.microsoft.com/en-us/aspnet/core/security/cross-site-scripting?view=aspnetcore-7.0#html-encoding-using-razor

jrmccannon:
Could we open a ticket to redo the page in one of our Vue app so that we can have a more reactive experience?

jonashendrickx:
> @jrmccannon Actually, me and Jonas had a conversation and reached this plan:
> 
> * Do a automatic page reload
> * On render of credentials, check if the credential was created recently (.Created < 10 minutes), if so add a small green new-label to the credential box

image

PR #195, opened 2023-11-13
Reorganize self-host directory

Tyrrrz:
- Renamed `docker` to `self-host` to avoid confusion with the other docker images
- Moved self-host-specific instructions to its own readme


Tyrrrz:
@jonashendrickx I might be missing some steps for the 'run' section of the instructions

jonashendrickx (approved):

PR #210, opened 2023-11-15
Enable null reference checks in `Service.csproj`

Tyrrrz:
This currently fails in tests due to how EF Core infers database schema based on NRT annotations. Need to assign proper nullability to all models in order for this error to go away.

Leaving this PR as a reminder that we need to do it.


Tyrrrz:
Just pinging for awareness (no action necessary): @jonashendrickx @jrmccannon

Tyrrrz:
Superseded by #340

PR #214, opened 2023-11-16
Exporting API keys no longer works

jonashendrickx:


abergs:
@jonashendrickx Do we know the cause of this?

jonashendrickx:
> @jonashendrickx Do we know the cause of this?

Nope, I went back far in the code, it wasn't working for quite a while.

PR #219, opened 2023-11-16
Manage plans using their SKU

jonashendrickx:


src/AdminConsole/Billing/Configuration/StripeOptions.cs
```diff
@@ -6,5 +6,7 @@ public class StripeOptions
 
     public string ApiKey { get; set; }
 
+    public ICollection<string> OnSale { get; set; }
```
abergs, line 9: nitpick: Perhaps use a different term, this signals to me that it's "on sale" as in 50% off.

src/AdminConsole/appsettings.json
```diff
@@ -36,7 +38,30 @@
...
+          "Features": [
+            "All Free features",
+            "Unlimited apps",
+            "Unlimited console administrators",
+            "Dedicated SLA",
+            "Prioritized support"
+          ]
+        }
+      },
+      "Pro-1": {
```
abergs, line 62: nitpick: Any reason for using this SKU instead of including the year? e.g. `PRO-2023-2`

abergs (approved):
Some small nitpicks which would be nice to adress, but not critical.

PR #220, opened 2023-11-16
Add `/signin/token` to create a verify token without the client flow

Tyrrrz:
JIRA: https://bitwarden.atlassian.net/browse/PAS-224?atlOrigin=eyJpIjoiOTczMDZjN2EwZmIyNDM2NDgyMGE5YzYwNWQzYzUwNmEiLCJwIjoiaiJ9

Request:

```
POST /signin/token

{
  "userId": "user123"
}
```

Response:

```
{
  "token": "verify_asdkasjkdlajscldjaslkdjckasljdc"
}
```

### DIFF
--- a/src/Service/Fido2ServiceEndpoints.cs
+++ b/src/Service/Fido2ServiceEndpoints.cs
@@ -252,16 +252,21 @@ public class Fido2ServiceEndpoints : IFido2Service
         return new TokenResponse(token);
     }
 
-    public Task<string> CreateSigninToken(SignInToken tokenProps)
+    public Task<string> CreateSigninToken(string userId)
     {
-        if (tokenProps.ExpiresAt == default)
+        ValidateUserId(userId);
+
+        _eventLogger.LogSigninTokenCreatedEvent(userId);
+
+        var tokenProps = new VerifySignInToken
         {
-            tokenProps.ExpiresAt = DateTime.UtcNow.AddSeconds(120);
-        }
-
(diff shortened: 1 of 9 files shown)


Tyrrrz:
This is super raw and MVP (would need to add more stuff for magic link support) but works for testing purposes:

image

Tyrrrz:
@abergs @jrmccannon if we can merge it before Monday (ideally before tomorrow), I'll be able to show more interesting tests on the demo :)

Tyrrrz:
Something to consider: what value of `type` should we use for this token?

Currently it's `manual_signin`. Potential ideas: `passkey_manual`, `passkey_magic`, `magic`, `virtual_signin`.

src/Service/Fido2ServiceEndpoints.cs
```diff
@@ -258,6 +252,27 @@ public async Task<TokenResponse> RegisterComplete(RegistrationCompleteDTO reques
         return new TokenResponse(token);
     }
 
+    public Task<string> CreateSigninToken(string userId)
+    {
+        ValidateUserId(userId);
```
jrmccannon, line 257: Not something that needs to be done now, but do we need to check to see if there's an existing credential? And then invalidate said credential?  That would be something to consider.

jrmccannon (approved):

Tyrrrz:
I'll merge it for now in this shape since it's only needed for demo and will still require a lot of tweaks for magic links.

The corresponding SDK part can wait.

PR #241, opened 2023-11-24
Authenticator Data

jonashendrickx:
We need to handpick the icons for the next bit


src/AdminConsole/AdminConsole.csproj
```diff
@@ -112,5 +112,12 @@
       <DependentUpon>_Credentials.cshtml</DependentUpon>
     </Compile>
   </ItemGroup>
+
+  <ItemGroup>
+    <Content Remove="Services\AuthenticatorData\authenticator-data.json" />
+    <None Include="Services\AuthenticatorData\authenticator-data.json">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
```
Tyrrrz, line 121:
```suggestion
  <ItemGroup>
    <None Include="Services\AuthenticatorData\authenticator-data.json" CopyToOutputDirectory="always" />
  </ItemGroup>

```

Tyrrrz (approved):

Tyrrrz:
Question: are we going to load the authenticator IDs once during deployments and seed them into the database? Or are we going to read them on startup and keep in memory? Or are we going to regularly pull new data in the background?

abergs:
> Question: are we going to load the authenticator IDs once during deployments and seed them into the database? Or are we going to read them on startup and keep in memory? Or are we going to regularly pull new data in the background?

I haven't looked at this properly, will do on Monday, but to answer your questions, my gut says:

I don't think we need to keep them in the DB. We could just make our lives simple and read the JSON on startup and keep a Dictionary in memory and have the `Credential.AuthenticatorName`-property lookup the `credential.aaGuid` in that dictionary. 

We don't have a need to dynamically add authenticators, so that would make it very few moving parts which would be easy to maintain I think?

Again - I haven't done a proper review yet :)

jonashendrickx:
> Question: are we going to load the authenticator IDs once during deployments and seed them into the database? Or are we going to read them on startup and keep in memory? Or are we going to regularly pull new data in the background?

It's an unofficial data source. Anything could happen to it any time. I was thinking of generating a .NET migrations script for it. Now it just parses the source once upon startup and updates the records, which is not great.

abergs:
> It's an unofficial data source. Anything could happen to it any time.

Do we have it as a local file in source control? If so, we can choose when to update and what goes in, instead of reading it hot out of github is my thought.

PR #242, opened 2023-11-24
Move Stripe API calls behind an interface and introduce a No-OP interface

abergs:
The goal of this PR is to enable running the AdminConsole "fully working" without Stripe, for testing Self-Host Beta.
The behaviour of the cloud environment should remain completely unchanged.

* Moves all stripe-api calls into it's own service, `SharedStripeBillingService`
* Introduces a base class, `BaseBillingService` for reacting to changes from the External billing provider (Stripe)
* Introduces a `NoopBillingService` that mocks a happy path for all payments, e.g. immediately upgrading stuff without charging.

We now have a 

Left todo:
* There are still some traces of Stripe in some Razor Pages.
* Some further renaming and cleaning up and doumentation.
* Test more in docker
* Cleanup testing values from `appsettings.json`
* Perhaps cleanup returning redirect URLs... feels dirty.

Related thoughts:
* Pedantic nitpick: _StripeOptions_ could be renamed to `BillingOptions` and include a `StripeApiKey`-field
* Those options could be moved to a separate settings file (?)


abergs:
@jonashendrickx This still need testing, but I think it's in OK shape. Just mentioning in case I will be on leave and it needs to be merged.

PR #247, opened 2023-11-27
Remove Expired Token Keys

jrmccannon:
There is an exception being thrown when a user has multiple expired token keys. Instead of looping the db call to delete, this just removes them all in one query.

Additional:
- This will also pass through the cancellation token from the endpoints that use the Fido2ServiceEndpoints
- Adds TimeProvider to ITenantProvider to be accessible in EfTenantStorage and DbTenantContext.


src/Service/TenantProvider.cs
```diff
@@ -5,4 +5,5 @@ namespace Passwordless.Service;
 public class TenantProvider : ITenantProvider
 {
     public string Tenant { get; set; }
+    public TimeProvider TimeProvider { get; set; }
```
jonashendrickx, line 8: Why add this here?

jonashendrickx (changes requested):

src/Service/Storage/Ef/EfTenantStorage.cs
```diff
@@ -196,6 +198,12 @@ public Task RemoveTokenKey(int keyId)
         return db.TokenKeys.Where(k => k.KeyId == keyId).ExecuteDeleteAsync();
     }
 
+    public Task RemoveExpiredTokenKeys(CancellationToken cancellationToken)
+    {
+        return db.TokenKeys.Where(x => (TimeProvider.GetUtcNow().DateTime - x.CreatedAt).TotalDays > 30)
```
jonashendrickx, line 203: Only inject TimeProvider here if you don't use it anywhere else. Don't litter it all over the code.

jonashendrickx (changes requested):

src/Service/Fido2ServiceEndpoints.cs
```diff
@@ -38,16 +38,15 @@ public class Fido2ServiceEndpoints : IFido2Service
         _eventLogger = eventLogger;
     }
 
-    private async Task Init()
+    private async Task Init(CancellationToken cancellationToken)
     {
-        await _tokenService.InitAsync();
+        await _tokenService.InitAsync(cancellationToken);
```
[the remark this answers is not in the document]
Tyrrrz: > This feels to me like a bit of a footgun (to share the same cancellationToken).

Why? 🤔

Tyrrrz (approved):

src/Service/Fido2ServiceEndpoints.cs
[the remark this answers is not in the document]
abergs: I may be wrong, but I've only ever seen cancellationToken being passed into the actual methods, not the "constructor" of the class.

We may of course pass cancellationToken to `InitAsync` to make sure we can abort anything we do in that method, but the way the code is implemented is that we set the cancellationToken as member of the class and it's utilized for all other methods of that class. That is a design pattern I haven't seen before and it can have benefits, but to me it seems the norm is to give the caller of a method the control to abort a single call. This is debatable, one could argue that it's a feature, but to me at least it's a new and surprising behaviour that would introduce severe bugs if the the lifetime of the class would be changed.

abergs (approved):

src/Service/Fido2ServiceEndpoints.cs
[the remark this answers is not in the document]
Tyrrrz: Yeah, storing cancellation tokens as fields is not _too_ unusual either, but it's not done here.

tests/Api.IntegrationTests/Behaviors/DeleteExpiredTokenKeysTests.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        // Act
+        _ = await _client.PostAsJsonAsync("register/token", tokenGenerator.Generate());
+
+        // Assert
+        var tokenKeys = await storage.GetTokenKeys();
+        tokenKeys.Should().NotBeNull();
+        tokenKeys.Any(x => x.CreatedAt < (DateTime.UtcNow.AddDays(-30))).Should().BeFalse();
+        tokenKeys.Any(x => x.CreatedAt >= (DateTime.UtcNow.AddDays(-30))).Should().BeTrue();
+    }
```
Tyrrrz, line 54: I would implement the test a bit differently:

1. Inject a fake `IDateTimeProvider` through the `WebApplicationFactory`
2. Simulate a behavior that would result in a new token being created (I lack the domain knowledge to know which one)
3. Move the time ahead in the future
4. Perform validations

In other words, we should not touch the database directly, but instead interact with the app through the means that are available to the user.

Also, I would not separate the test out like this, since it has to be tied out to some surface-level behavior (not necessarily endpoint) that's observable by the user, right?

jrmccannon: Good point. I'll make the change tomorrow morning...or later tonight if its really bugging me.

jrmccannon: > Also, I would not separate the test out like this, since it has to be tied out to some surface-level behavior (not necessarily endpoint) that's observable by the user, right?

Any endpoint that instantiates the `TokenServicce` and therefore the `Fido2ServiceEndpoints` creates this. Basically all the endpoints that return a token.  If I think about wholistic coverage, I'd have to replicate this test for each endpoint that returned a token...which seems like a lot of duplication.  Where would I put a test like this? Or would I simply do what I need to for the endpoint that is currently causing the error in production `signin/begin`?

Tyrrrz: > Or would I simply do what I need to for the endpoint that is currently causing the error in production signin/begin?

I would do this. I assume most of the code you cover with this test is reused across other endpoints? In which case, you will not need to duplicate the test for each endpoint.

We have coverage maps now, so we can see where exactly we lack coverage – this way we don't have to approach testing like a numbers game but go off of what needs the most attention.

Tyrrrz: In general, I think it's a decent approach to add tests only if we find a bug that could have been uncovered with that test. No need to overthink things too much.

jonashendrickx (approved):

PR #263, opened 2023-11-28
Add Codecov

Tyrrrz:
Doesn't work currently because we need IT to add secrets

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -133,8 +133,6 @@ jobs:
 
       - name: Upload coverage
         uses: codecov/codecov-action@eaaf4bedf32dbdc6b720b63067d99c4d77d6047d # 3.1.4
-        with:
-          token: ${{ secrets.CODECOV_TOKEN }}
 
   # Publish the apps
   pack:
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -127,6 +131,11 @@ jobs:
           reporter: dotnet-trx
           fail-on-error: true
 
+      - name: Upload coverage
+        uses: codecov/codecov-action@eaaf4bedf32dbdc6b720b63067d99c4d77d6047d # 3.1.4
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
Tyrrrz, line 137:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -80,7 +80,7 @@ jobs:
         uses: actions/setup-dotnet@3447fd6a9f9e57506b15f895c5b76d3b197dc7c2 # 3.2.0
         with:
           dotnet-version: 8.0.x
```
withinfocus, line 82: 🌱 Since you're using `global.json` none of your usages of this Action need to specify a .NET version and it'll read that file on its own. Easier to maintain.

PR #285, opened 2023-12-03
Refactor Fido2ServiceEndpoints & TokenService

jonashendrickx:


Tyrrrz (approved):

src/Service/TokenService.cs
```diff
@@ -237,4 +207,39 @@ private bool VerifyMac(Key key, byte[] data, byte[] mac)
...
+                    await _storage.RemoveExpiredTokenKeys(CancellationToken.None);
+                }
+                catch (Exception)
+                {
+                    _log.LogError("Failed to remove old key, account={accountName}", _tenant);
+                }
+            }
+        }
+
+        return _alternatives;
```
Tyrrrz, line 243: There's a potential of a race condition if `GetAlternativeAsync` is called twice in quick succession while `_alternatives` is null.

It might be better to initialize a local `var alternatives`, perform all manipulations, and only set `_alternative = alternatives` in the return statement

PR #286, opened 2023-12-04
[PAS64] Limit the creation of users to 10k on free plan

jonashendrickx:
image

image


Tyrrrz:
@jonashendrickx did you merge this while actively participating in a call? 😂

PR #293, opened 2023-12-07
[DEVOPS-1658] Setup DCT for and publish Passwordless images to DockerHub

michalchecinski:


.github/workflows/docker.yml
```diff
@@ -123,14 +123,17 @@ jobs:
...
 
-      - name: Login to the registry
-        run: >
-          echo ${{ secrets.GHCR_TOKEN }} |
-          docker login ${{ env.CONTAINER_NAMESPACE }}
-          --username abergs
-          --password-stdin
+      - name: Setup DCT
+        id: setup-dct
+        uses: bitwarden/gh-actions/setup-docker-trust@main
```
Tyrrrz, line 128: Should this be snapshotted to a specific verison of the action?

.github/workflows/docker.yml
```diff
@@ -141,3 +144,8 @@ jobs:
           ${{ github.event_name == 'release' && format('--tag {0}/{1}:stable', env.CONTAINER_NAMESPACE, matrix.name) || '' }}
           ${{ github.event_name == 'release' && format('--tag {0}/{1}:{2}', env.CONTAINER_NAMESPACE, matrix.name, github.ref_name) || '' }}
           ${{ inputs.image-tag && format('--tag {0}/{1}:{2}', env.CONTAINER_NAMESPACE, matrix.name, inputs.image-tag) || '' }}
+
+      - name: Log out of Docker and disable Docker Notary
+        run: |
+          docker logout
+          echo "DOCKER_CONTENT_TRUST=0" >> $GITHUB_ENV
```
Tyrrrz, line 151: Does this part matter, considering the agents don't retain their state between runs?

.github/workflows/docker.yml
michalchecinski: We pin our own actions to `main` to avoid the burden of updating them manually. We don't use tags, so we had problems using renovate for this task.

.github/workflows/docker.yml
michalchecinski: In this particular case it doesn't matter. But we tend to include it for future if more steps will be added after this 😃

Tyrrrz: Makes sense. Do you think this can also be distributed as a bitwarden action then, to make things cleaner? 🙂

Tyrrrz (approved):

Tyrrrz:
Looks good overall!

Just to confirm, with this PR merged, we will have out images published to Docker Hub, correct? Where will they be available to pull from?

.github/workflows/docker.yml
```diff
@@ -123,14 +123,17 @@ jobs:
       - name: Install Docker Buildx
         uses: docker/setup-buildx-action@f95db51fddba0c2d1ec667646a06c2ce06100226 # v3.0.0
 
-      - name: Login to the registry
-        run: >
-          echo ${{ secrets.GHCR_TOKEN }} |
-          docker login ${{ env.CONTAINER_NAMESPACE }}
-          --username abergs
-          --password-stdin
+      - name: Setup DCT
```
Tyrrrz, line 126: To make things clearer:

```suggestion
      - name: Setup Docker Content Trust
```

.github/workflows/docker.yml
michalchecinski: It can be distributed as a bitwarden action. But it's so small that I don't think it makes sense to make them as reusable action.

michalchecinski:
@Tyrrrz That is correct. They'll be available on public bitwarden dockerhub here: 
- https://hub.docker.com/r/bitwarden/passwordless-self-host
- https://hub.docker.com/r/bitwarden/passwordless-test-api

Tyrrrz:
Thanks!

PR #298, opened 2023-12-11
Added user check for generate token endpoint.

jrmccannon:
We needed to validate that the user exists before issuing a token for sign-in. This will check Aliases and Credentials.

### DIFF
--- a/src/Api/Endpoints/Signin.cs
+++ b/src/Api/Endpoints/Signin.cs
@@ -8,18 +8,16 @@ namespace Passwordless.Api.Endpoints;
 
 public static class SigninEndpoints
 {
-    private record SigninTokenRequest(string UserId);
-
-    private record SigninTokenResponse(string Token);
+    public record SigninTokenResponse(string Token);
 
     public static void MapSigninEndpoints(this WebApplication app)
     {
         app.MapPost("/signin/generate-token", async (
-                SigninTokenRequest signinToken,
+                SignInTokenRequest signinToken,
                 IFido2Service fido2Service
             ) =>
             {
-                var result = await fido2Service.CreateSigninToken(signinToken.UserId);
+                var result = await fido2Service.CreateSigninToken(signinToken);
 
                 return Ok(new SigninTokenResponse(result));
             })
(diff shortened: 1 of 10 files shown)


src/Api/Endpoints/Signin.cs
```diff
@@ -8,18 +8,16 @@ namespace Passwordless.Api.Endpoints;
...
-    private record SigninTokenRequest(string UserId);
-
-    private record SigninTokenResponse(string Token);
+    public record SigninTokenResponse(string Token);
 
     public static void MapSigninEndpoints(this WebApplication app)
     {
         app.MapPost("/signin/generate-token", async (
-                SigninTokenRequest signinToken,
+                SignInTokenRequest signinToken,
```
Tyrrrz, line 16: Nit

Seems like we're slightly inconsistent on the casing of `Signin`/`SignIn`. I thought, when used as a noun, `Signin` is more common. But either way, we should probably stick to one convention.

src/Service/Models/SignInTokenRequest.cs
```diff
@@ -0,0 +1,6 @@
+namespace Passwordless.Service.Models;
+
+public class SignInTokenRequest(string userId) : RequestBase
```
Tyrrrz, line 3: Should this be a record maybe?

Tyrrrz (approved):

src/Service/Models/SignInTokenRequest.cs
jrmccannon: I added `RequestBase` which is a class.  `RequestBase` has `Origin` and `RpId` which is on the other request endpoints.  Since RequestBase is a class, the request object has to be a class. That was the reason I switched it over.

Tyrrrz: Can `RequestBase` be a record then too? 🤔 

This is me nitpicking though

jrmccannon: No worries about the nits. Glad we're talking it through. If we change the other request objects to records too, we can change request base to a record. Let's hold off though until this is merged in and then make that a separate pr.

PR #302, opened 2023-12-13
Enforcement of scopes

jonashendrickx:


src/Api/Authorization/Extensions.cs
```diff
@@ -19,6 +21,17 @@ public static RouteHandlerBuilder RequirePublicKey(this RouteHandlerBuilder buil
         return builder.RequireAuthorization(Constants.PublicKeyPolicy);
     }
 
+    public static RouteHandlerBuilder RequireAuthorization(this RouteHandlerBuilder builder, ApiKeyTypes type, string scope)
```
Tyrrrz, line 24: Could the scope also be an enum?

Tyrrrz (approved):

PR #304, opened 2023-12-13
Migrate Api Key Manangement tests

jrmccannon:
This moves the Api Key Management Api tests out of the endpoint test file to the new Integration tests project.  Also includes an addition to the Database seed method to add scopes to the public key that's created on startup.


src/Service/EventLog/Models/EventResponse.cs
```diff
@@ -4,6 +4,8 @@ namespace Passwordless.Service.EventLog.Models;
 
 public class EventResponse
 {
+    public EventResponse() { }
+
     public EventResponse(DateTime performedAt, string message, string performedBy, string applicationId, EventType eventType, Severity severity, string subject, string apiKeyId)
```
Tyrrrz, line 9: If we already have `init` props, do we need the constructor at all?

tests/Api.IntegrationTests/Helpers/App/CreateAppHelpers.cs
```diff
@@ -18,4 +20,7 @@ public static async Task<HttpResponseMessage> CreateApplication(this HttpClient
 
         return await client.PostAsJsonAsync($"/admin/apps/{applicationName}/create", AppCreateGenerator.Generate());
     }
+
+    public static Task<HttpResponseMessage> CreateApplication(this HttpClient client)
```
Tyrrrz, line 24: Should these method be suffixed by `Async`?

Tyrrrz (approved):
Looks great :)

src/Service/EventLog/Models/EventResponse.cs
jrmccannon: The JsonSerializer was complaining about not being able to use the constructor.

PR #325, opened 2023-12-26
Self-hosting not working on reverse proxy with port =! 5701

jonashendrickx:


Tyrrrz (dismissed):

self-host/traefik/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+    image: ghcr.io/passwordless/passwordless-self-host:latest
+    volumes:
+      - /Users/jonashendrickx/src/bitwarden/passwordless-server/self-host/traefik/pwdcache:/etc/bitwarden_passwordless
+    restart: unless-stopped
+    environment:
+      # The port to which our app will listen publicly, necessary for non-request contexts.
+      BWP_PORT: 80
+      BWP_DOMAIN: passwordless.docker.localhost
+    labels:
+      - "traefik.http.routers.passwordless.rule=Host(`passwordless.docker.localhost`)"
```
Tyrrrz, line 29: Is this file included as an example or is it used for something?

jonashendrickx: Yeah, it's an example config for use with Traefik as a reverse proxy.

self-host/traefik/simple-example/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
+version: '3'
```
Tyrrrz, line 1: What about `self-host/examples/traefik/docker-compose.yml` instead?

Tyrrrz (dismissed):

self-host/traefik/simple-example/docker-compose.yml
jonashendrickx: Moved it

Tyrrrz (approved):

PR #331, opened 2023-12-28
Simple reporting: Total Users/Credentials charts

jonashendrickx:


src/AdminConsole/AdminConsole.csproj
```diff
@@ -55,5 +55,13 @@
   <Target Name="Run NPM Build" AfterTargets="PostBuildEvent">
     <Exec Command="npm install" />
     <Exec Command="npm run build" />
+
```
jonashendrickx, line 58: This will copy the necessary files from node_modules to our statically served assets folder for getting charts to work.

src/AdminConsole/Components/Pages/App/ReportingComponents/TotalCredentialsCountChart.razor
```diff
@@ -0,0 +1,18 @@
+@using Passwordless.Common.Models.Reporting
+<LineChart Id="total-credentials-count-chart" XCategories="@XCategories" Series="@Values"></LineChart>
+
+@code {
+    [CascadingParameter]
```
jonashendrickx, line 5: We have to pass the `Task`, as the data isn't initialized yet, we cannot await in our parent component.

src/Common/BasePeriodicBackgroundService.cs
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    protected CancellationTokenSource CancellationToken { get; } = new();
+
+    protected override Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        // Calculate the time until the next execution
+        var currentTime = _timeProvider.GetLocalNow().TimeOfDay;
+
+        TimeSpan initialDelay;
```
jonashendrickx, line 44: This will execute the reporting always at the same time during the day.

src/AdminConsole/Components/Pages/App/Reporting.razor
```diff
@@ -0,0 +1,67 @@
+@page "/app/{app}/Reporting"
```
jonashendrickx, line 1: We don't need to modify the route in the middleware anymore. Do not rename {app}.

src/AdminConsole/Components/Shared/Charts/LineChart.razor
```diff
@@ -0,0 +1,47 @@
+@attribute [CascadingTypeParameter(nameof(TXCategory))]
+@typeparam TXCategory
+
+@inject IHttpContextAccessor HttpContextAccessor
+
+<div id="@Id"></div>
```
jonashendrickx, line 6: Each chart has a unique id, so we can render it accordingly.

src/Service/Storage/Ef/DbTenantFactory.cs
```diff
@@ -19,6 +20,14 @@ public EfTenantStorageFactory(IServiceProvider services, TimeProvider timeProvid
         _timeProvider = timeProvider;
     }
 
+    public ITenantStorage Create()
```
jonashendrickx, line 23: We do not need to pass the `appId`, and leverage dependency injection.

jrmccannon (dismissed):

jrmccannon (approved):

src/AdminConsole/Components/App.razor
```diff
@@ -5,14 +5,15 @@
     <meta charset="utf-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
     <base href="/"/>
-    <link rel="stylesheet" href="css/site.css" asp-append-version="true"/>
-    <link rel="stylesheet" href="css/tailwind.css" asp-append-version="true"/>
+    <link rel="stylesheet" href="css/site.css" />
+    <link rel="stylesheet" href="css/tailwind.css" />
     <HeadOutlet/>
```
abergs, line 10: Put this back :)

src/AdminConsole/Components/App.razor
```diff
@@ -5,14 +5,15 @@
...
 </head>
 
 <body class="h-full bg-gray-50">
     <Routes/>
     <script src="_framework/blazor.web.js"></script>
-    <script type="text/javascript" src="js/core.js" asp-append-version="true"></script>
+    <script type="text/javascript" src="lib/apexcharts/apexcharts.min.js"></script>
+    <script type="text/javascript" src="js/core.js"></script>
```
abergs, line 17: Add asp-append-version

jonashendrickx: @abergs It's done. asp-append-version is not supported.

src/Api/Endpoints/Reporting.cs
```diff
@@ -0,0 +1,26 @@
...
+using Passwordless.Service;
+using static Microsoft.AspNetCore.Http.Results;
+
+namespace Passwordless.Api.Endpoints;
+
+public static class ReportingEndpoints
+{
+    public static void MapReportingEndpoints(this WebApplication app)
+    {
+        var group = app.MapGroup("/apps/{appId}/reporting")
```
abergs, line 12: Add note about this being an internal endpoint (should not be documented or expected to be stable)

src/Api/Helpers/DesignTimeFactories.cs
```diff
@@ -1,26 +0,0 @@
...
-
-public class MsSqlContextFactory : IDesignTimeDbContextFactory<DbGlobalMsSqlContext>
-{
-    public DbGlobalMsSqlContext CreateDbContext(string[] args)
-    {
-        var options = new DbContextOptionsBuilder<DbGlobalMsSqlContext>();
-        options.UseSqlServer(args.Length == 1 ? args[0] : null);
-        return new DbGlobalMsSqlContext(options.Options);
-    }
-}
```
abergs, line 26: @jonashendrickx Why are these deleted? Are they not needed because of this PR?

src/Service/Models/PeriodicCredentialReport.cs
```diff
@@ -0,0 +1,19 @@
+namespace Passwordless.Service.Models;
+
+public class PeriodicCredentialReport : PerTenant
+{
+    public DateTime CreatedAt { get; set; }
```
abergs, line 5: Should we perhaps tweak this property to always return midnight so that we never accidently include current time (which would result in double reporting)

abergs: Or perhaps: https://learn.microsoft.com/en-us/dotnet/standard/datetime/how-to-use-dateonly-timeonly

src/Service/Storage/Ef/DbGlobalContext.cs
```diff
@@ -11,8 +11,13 @@ namespace Passwordless.Service.Storage.Ef;
...
 public abstract class DbGlobalContext : DbContext
 {
-    public DbGlobalContext(DbContextOptions options) : base(options)
+    protected readonly TimeProvider TimeProvider;
+
+    protected DbGlobalContext(
+        DbContextOptions options,
+        TimeProvider timeProvider) : base(options)
     {
+        TimeProvider = timeProvider;
```
abergs, line 20: Is TimeProvider used in this class?

src/Service/Models/PeriodicCredentialReport.cs
jonashendrickx: I think I'm already doing this a layer up or so.

jonashendrickx: `public async Task<int> UpdatePeriodicCredentialReportsAsync()
    {
        var result = _db.AccountInfo
            .GroupJoin(
                _db.Credentials,
                accountInfo => accountInfo.Tenant,
                credential => credential.Tenant,
                (accountInformation, credentials) => new PeriodicCredentialReport
                {
                    Tenant = accountInformation.Tenant,
                    UsersCount = credentials.Select(x => x.UserId).Distinct().Count(),
                    CredentialsCount = credentials.Count(),
                    CreatedAt = _timeProvider.GetUtcNow().UtcDateTime.Date // here
                });

        await _db.PeriodicCredentialReports.AddRangeAsync(result);

        var rows = await _db.SaveChangesAsync();

        return rows;
    }`

src/AdminConsole/Components/Shared/BlazorStylesheet.razor
```diff
@@ -0,0 +1,21 @@
+@inject IFileVersionProvider Version
+
+<link rel="stylesheet" href="@Href" />
```
abergs, line 3: @jonashendrickx Remind me, what did we say about the cachebusting? Will we need to fix it in a separate PR?

src/Api/Endpoints/Reporting.cs
abergs: Absolutely, but we should do a proper API review before making it public since it's hard to change it afterwards.

src/Service/Migrations/Mssql/20240108104909_AddTablePeriodicCredentialReports.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    /// <inheritdoc />
+    protected override void Up(MigrationBuilder migrationBuilder)
+    {
+        migrationBuilder.CreateTable(
+            name: "PeriodicCredentialReports",
+            columns: table => new
+            {
+                Tenant = table.Column<string>(type: "nvarchar(450)", nullable: false),
+                CreatedAt = table.Column<DateTime>(type: "datetime2", nullable: false),
```
abergs, line 18: @jonashendrickx Did you re-generate these after adding DateOnly? I would expect this to be date in sqlserver 👀

src/Service/Storage/Ef/DbGlobalContext.cs
```diff
@@ -77,6 +78,18 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
                 .IsRequired();
         });
 
+        modelBuilder.Entity<PeriodicCredentialReport>(builder =>
+        {
+            builder.HasKey(x => new { x.Tenant, x.CreatedAt });
+
+            builder.Property(x => x.CreatedAt);
```
abergs, line 85: nit: Is this line a no-op or does it do something?

abergs (approved):

PR #334, opened 2024-01-04
Adding rate limiting to Admins.cshtml.cs

jrmccannon:
This adds rate limiting to the Admins page to address ability to spam emails using the invite admin form.

The settings used are 
```
            options.PermitLimit = 10;
            options.Window = TimeSpan.FromMinutes(1);
            options.QueueLimit = 0;
```

This means we limit 10 requests to the resource for 1 min. The `QueueLimit` determines the number of requests we'll hold in queue to execute later once the limit has been reached. This was set to 0 to drop all requests once the limit is reached.

I had to apply the rate limit attribute to the entire razor page due to a [limitation of the RateLimitMiddleware](https://learn.microsoft.com/en-us/aspnet/core/performance/rate-limit?view=aspnetcore-8.0#applying-attributes-to-razor-pages).

In order to also reduce the ability of a user to abuse this page, I enforced a unique restriction on the email the invite is being sent to.  If the email already exists in the invite list, we won't send the invite. They'll have to cancel the old invite first.

### DIFF
--- a/src/AdminConsole/Pages/Organization/Admins.cshtml.cs
+++ b/src/AdminConsole/Pages/Organization/Admins.cshtml.cs
@@ -3,6 +3,7 @@ using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.AspNetCore.RateLimiting;
 using Passwordless.AdminConsole.EventLog.Loggers;
 using Passwordless.AdminConsole.Helpers;
 using Passwordless.AdminConsole.Identity;
@@ -10,6 +11,7 @@ using Passwordless.AdminConsole.Services;
 
 namespace Passwordless.AdminConsole.Pages.Organization;
 
+[EnableRateLimiting("fixed")]
 public class Admins : PageModel
 {
     private readonly IDataService _dataService;
@@ -108,10 +110,15 @@ public class Admins : PageModel
         var orgId = org.Id;
         var orgName = org.Name;
         ConsoleAdmin user = await _dataService.GetUserAsync();
-        var userEmail = user.Email;
+        var userEmail = user.Email!;
         var userName = user.Name;
 
-        await _invitationService.SendInviteAsync(form.Email, orgId, orgName, userEmail, userName);
+        var existingInvites = await _invitationService.GetInvitesAsync(User.GetOrgId()!.Value);
+
+        if (existingInvites.All(x => x.ToEmail != form.Email))
+        {
+            await _invitationService.SendInviteAsync(form.Email, orgId, orgName, userEmail, userName);
+        }
 
         _eventLogger.LogInviteAdminEvent(user, form.Email, _systemClock.UtcNow.UtcDateTime);
 
--- a/src/AdminConsole/Program.cs
+++ b/src/AdminConsole/Program.cs
@@ -1,6 +1,8 @@
+using System.Threading.RateLimiting;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Components.Authorization;
 using Microsoft.AspNetCore.Mvc.Infrastructure;
+using Microsoft.AspNetCore.RateLimiting;
 using Microsoft.Extensions.Options;
 using Passwordless.AdminConsole;
 using Passwordless.AdminConsole.Authorization;
@@ -155,6 +157,15 @@ void RunTheApp()
 
     builder.Services.AddAntiforgery();
 
+    builder.Services.AddRateLimiter(limiter => limiter
+        .AddFixedWindowLimiter(policyName: "fixed", options =>
+        {
+            options.PermitLimit = 10;
+            options.Window = TimeSpan.FromMinutes(1);
+            options.QueueProcessingOrder = QueueProcessingOrder.OldestFirst;
+            options.QueueLimit = 0;
+        }));
+
     WebApplication app;
     try
     {
@@ -192,6 +203,7 @@ void RunTheApp()
     app.UseStaticFiles();
     app.UseSerilogRequestLogging();
     app.UseRouting();
+    app.UseRateLimiter();
     app.UseAntiforgery();
     app.MapHealthEndpoints();
     app.UseAuthentication();
--- a/src/AdminConsole/Services/IInvitationService.cs
+++ b/src/AdminConsole/Services/IInvitationService.cs
@@ -4,7 +4,7 @@ namespace Passwordless.AdminConsole.Services;
 
 public interface IInvitationService
 {
-    Task SendInviteAsync(string toEmail, int TargetOrgId, string TargetOrgName, string fromEmail, string fromName);
+    Task SendInviteAsync(string toEmail, int targetOrgId, string targetOrgName, string fromEmail, string fromName);
     Task<List<Invite>> GetInvitesAsync(int orgId);
     Task CancelInviteAsync(string hashedCode);
     Task<Invite> GetInviteFromRawCodeAsync(string code);
--- a/src/AdminConsole/Services/InvitationService.cs
+++ b/src/AdminConsole/Services/InvitationService.cs
@@ -24,13 +24,13 @@ public class InvitationService<TDbContext> : IInvitationService where TDbContext
         _actionContextAccessor = actionContextAccessor;
     }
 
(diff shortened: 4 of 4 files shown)


src/AdminConsole/Pages/Organization/Admins.cshtml.cs
```diff
@@ -3,13 +3,15 @@
...
 using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.AspNetCore.RateLimiting;
 using Passwordless.AdminConsole.EventLog.Loggers;
 using Passwordless.AdminConsole.Helpers;
 using Passwordless.AdminConsole.Identity;
 using Passwordless.AdminConsole.Services;
 
 namespace Passwordless.AdminConsole.Pages.Organization;
 
+[EnableRateLimiting("fixed")]
```
jonashendrickx, line 14: Remove the rate limiting on page level, instead move OnPostInvite to an endpoint, and have the form submission calling the endpoint.

Then apply the rate limiting to the endpoint.

`app.MapGet("/", () => Results.Ok($"Fixed Window Limiter {GetTicks()}"))
                           .RequireRateLimiting(fixedPolicy);`
                           
Consider also using a static readonly representation of the rate limiting policy.

jonashendrickx (changes requested):

src/AdminConsole/Pages/Organization/Admins.cshtml.cs
jrmccannon: Chatted with @abergs about splitting this out into a separate API endpoint. He felt it wasn't worth moving it out of the page.  We decided upping the limit to 100 and put a few more safe guards in.  Only allowing 1 invite per email at a time and 10 pending invites at a time.

Also went ahead with the static definition of the policy and adding of rate limiting.

jonashendrickx: Can you not use something better, i.e.:

1. Preventing double submissions in the first place
2. Upon generating links, I suppose they are valid for a limited time, so you can only have up to 3 active magic links for a given user or admin.

src/AdminConsole/Pages/Organization/Admins.cshtml.cs
```diff
@@ -92,28 +94,39 @@ public async Task<IActionResult> OnPostDelete(string userId)
...
     {
         CanInviteAdmin = await _dataService.CanInviteAdminAsync();
+
         if (CanInviteAdmin is false)
         {
             ModelState.AddModelError("error", "You need to upgrade to a paid organization to invite more admins.");
             return await OnGet();
         }
 
+        Models.Organization org = await _dataService.GetOrganizationAsync();
```
Tyrrrz, line 104:
```suggestion
        var org = await _dataService.GetOrganizationAsync();
```

src/AdminConsole/Pages/Organization/Admins.cshtml.cs
```diff
@@ -92,28 +94,39 @@ public async Task<IActionResult> OnPostDelete(string userId)
...
+        var existingInvites = await _invitationService.GetInvitesAsync(org.Id);
+
+        if (existingInvites.Count == 10)
```
Tyrrrz, line 107: Just in case

```suggestion
        if (existingInvites.Count >= 10)
```

src/AdminConsole/Pages/Organization/Admins.cshtml.cs
```diff
@@ -92,28 +94,39 @@ public async Task<IActionResult> OnPostDelete(string userId)
...
         if (!ModelState.IsValid)
         {
             // todo: Is there a pattern where we don't need to repeat this?
             return await OnGet();
         }
 
-        Models.Organization org = await _dataService.GetOrganizationAsync();
-        var orgId = org.Id;
-        var orgName = org.Name;
         ConsoleAdmin user = await _dataService.GetUserAsync();
```
Tyrrrz, line 125:
```suggestion
        var user = await _dataService.GetUserAsync();
```

src/AdminConsole/Pages/Organization/Admins.cshtml.cs
```diff
@@ -92,28 +94,39 @@ public async Task<IActionResult> OnPostDelete(string userId)
...
+        Models.Organization org = await _dataService.GetOrganizationAsync();
+        var existingInvites = await _invitationService.GetInvitesAsync(org.Id);
+
+        if (existingInvites.Count == 10)
+        {
+            ModelState.AddModelError("error", "You can only have 10 pending invites at a time.");
+            return await OnGet();
+        }
+
+        if (existingInvites.Any(x => x.ToEmail == form.Email))
```
Tyrrrz, line 113: Note that it might not catch all duplicates, including emails that use `+` or those written in a different case. If this is something that people may abuse, then maybe we need a more resilient check?

src/AdminConsole/RateLimiting/AdminPageRateLimit.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public const string PolicyName = "organizationIdFixedLength";
+    private const int PermitLimit = 100;
+    private static readonly TimeSpan Window = TimeSpan.FromMinutes(1);
+    private const int QueueLimit = 0;
+
+    public static RateLimiterOptions AddAdminPageRateLimitPolicy(this RateLimiterOptions limiter) =>
+        limiter.AddPolicy(PolicyName, context =>
+            RateLimitPartition.GetFixedWindowLimiter(
+                context.User.GetOrgId()!.Value.ToString(),
```
Tyrrrz, line 17: When can `.GetOrgId()` return null?

jrmccannon: If the user does not have the "OrgId" claim, which would occur if the user is unauthenticated (in which they would be redirected to the login page) or if they were never added to an org (which is very highly unlikely). 

TLDR: it shouldn't happen.

src/AdminConsole/Pages/Organization/Admins.cshtml.cs
abergs: good observation, in this case we just want a simple check. We should probably to a case-insensitive comparison.

abergs (approved):

abergs:
Approved, see comments from Oleksii

abergs (dismissed):

src/AdminConsole/Pages/Organization/Admins.cshtml.cs
```diff
@@ -92,28 +94,39 @@ public async Task<IActionResult> OnPostDelete(string userId)
...
+        var org = await _dataService.GetOrganizationAsync();
+        var existingInvites = await _invitationService.GetInvitesAsync(org.Id);
+
+        if (existingInvites.Count >= 10)
+        {
+            ModelState.AddModelError("error", "You can only have 10 pending invites at a time.");
+            return await OnGet();
+        }
+
+        if (existingInvites.Any(x => string.Equals(x.ToEmail, form.Email, StringComparison.CurrentCultureIgnoreCase)))
```
Tyrrrz, line 113: I don't think non-ascii characters can be used in emails, can they?

```suggestion
        if (existingInvites.Any(x => string.Equals(x.ToEmail, form.Email, StringComparison.OrdinalIgnoreCase)))
```

src/AdminConsole/Pages/Organization/Admins.cshtml.cs
```diff
@@ -92,28 +94,39 @@ public async Task<IActionResult> OnPostDelete(string userId)
...
-        ConsoleAdmin user = await _dataService.GetUserAsync();
-        var userEmail = user.Email;
-        var userName = user.Name;
+        var user = await _dataService.GetUserAsync();
 
-        await _invitationService.SendInviteAsync(form.Email, orgId, orgName, userEmail, userName);
+        await _invitationService.SendInviteAsync(form.Email, org.Id, org.Name, user.Email!, user.Name);
 
-        _eventLogger.LogInviteAdminEvent(user, form.Email, _systemClock.UtcNow.UtcDateTime);
+        _eventLogger.LogInviteAdminEvent(user, form.Email, _timeProvider.GetUtcNow().UtcDateTime);
```
Tyrrrz, line 129: Nit/future: Maybe `TimeProvider` can be injected in some kind of context inside the event logger so that `_timeProvider.GetUtcNow().UtcDateTime` can be odne implicitly?

Tyrrrz (approved):

Tyrrrz (approved):

PR #341, opened 2024-01-10
Route application logs through tests

Tyrrrz:


jonashendrickx (dismissed):

jrmccannon (approved):

Tyrrrz:
This took a lot more work because I couldn't find a nice way to get things plumbed nicely without too many changes. Also, the fact that we were using Serilog required a workaround as well.

Before: 
image

After: 
image

@abergs need your review because merging is blocked.

src/Api/Program.cs
```diff
@@ -29,35 +29,40 @@
...
+                    configuration: ddConfig);
+            }
         }
-    }
-});
+    },
+    false,
+    // Pass log events down to other logging providers (e.g. Microsoft) after Serilog, so
+    // that they can be processed in a uniform way in tests.
+    true
```
Tyrrrz, line 64: This was required because Serilog overshadows other logging providers and there's no easy way (that I found) to append another sink to an existing Serilog configuration for use in tests.

abergs (approved):
Complex stuff! Happy you got it working, although my gut feeling says it's not the last time we'll be wrangling this code. Let's hope I'm wrong 👍

PR #342, opened 2024-01-10
Fix Enable/Disable of the Manual Verification Token Endpoint

jrmccannon:
All fixed up. Thanks @abergs

### DIFF
--- a/src/AdminConsole/Pages/App/Settings/Settings.cshtml.cs
+++ b/src/AdminConsole/Pages/App/Settings/Settings.cshtml.cs
@@ -206,7 +206,9 @@ public class SettingsModel : BaseExtendedPageModel
 
         try
         {
-            if (Convert.ToBoolean(Request.Form[ManualVerificationTokenCheckboxId].FirstOrDefault()))
+            Request.Form.TryGetValue(ManualVerificationTokenCheckboxId, out var values);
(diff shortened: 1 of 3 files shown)


src/AdminConsole/Pages/App/Settings/Settings.cshtml.cs
```diff
@@ -206,7 +206,9 @@ public async Task<IActionResult> OnPostSettingsAsync()
 
         try
         {
-            if (Convert.ToBoolean(Request.Form[ManualVerificationTokenCheckboxId].FirstOrDefault()))
+            Request.Form.TryGetValue(ManualVerificationTokenCheckboxId, out var values);
```
jonashendrickx, line 209: Why is it not getting it from the form's model directly/

jrmccannon: For some reason the bindings aren't working on that page. Do you see something that's missing from the previous version of the page?

abergs:
I commited a small test - this seemed to be working for me. Can you verify?

src/AdminConsole/Pages/App/Settings/Settings.cshtml.cs
jrmccannon: Fixed!

abergs (approved):

PR #347, opened 2024-01-11
Fix HTTP status 500.

jonashendrickx:


src/Api/Middleware/EventLogContextMiddleware.cs
```diff
@@ -13,7 +13,14 @@ public class EventLogContextMiddleware
 
     public async Task InvokeAsync(HttpContext context, IEventLogContext eventLogContext, IFeatureContextProvider featureContextProvider)
     {
+        if (!context.Items.ContainsKey("__AuthorizationMiddlewareWithEndpointInvoked"))
```
abergs, line 16: Would benefit from a comment why this is done. Is there any other "public" way to check if this is an authorized endpoint, `__AuthorizationMiddlewareWithEndpointInvoked` feels like a magic internal string. But I may be wrong?

jonashendrickx: @abergs  see comment :-) !

src/Api/Middleware/EventLogStorageCommitMiddleware.cs
```diff
@@ -28,7 +29,11 @@ public async Task InvokeAsync(HttpContext context, IServiceProvider provider)
         {
             try
             {
-                await provider.GetRequiredService<IEventLogger>().FlushAsync();
+                var eventLoggerContext = provider.GetRequiredService<IEventLogContext>();
+                if (!string.IsNullOrEmpty(eventLoggerContext.TenantId))
```
jrmccannon, line 33: This breaks all event logging from an unauthenticated standpoint. I wrote some tests to validate those instances and can add them to this branch.

jrmccannon (changes requested):
The requirement of the eventLoggerContext to exist won't work because it isn't invoked when the user hasn't been authenticated yet.

PR #348, opened 2024-01-12
Use npm ci

abergs:
This PR changes the dotnetbuild to use npm ci so that the install is clean, without updating lock files.

This causes a dependency on npm > 10. Not sure if that needs to be documented:
* `nvm use node`
* `npm install -g npm`

This PR also removes the slow/needles use of npx.


jonashendrickx (approved):

Tyrrrz:
By the way, we seem to have both `yarn.lock` and `package-lock.json`. Should we standardize on one package manager? As far as I know, they don't work very well when used side-by-side.

abergs:
yeah, let's nuke yarn?

PR #354, opened 2024-01-13
Fix event logging context bug on non-existing endpoints

jonashendrickx:
Fixes 2 separate things:

- Endpoints that we haven't defined, shouldn't execute event logging middleware, i.e. '/trollz?api=123'
- While event logging is also capturing invalid api keys and api secrets, if that are not in the expected format, then we would be able to define the tenant context we'd have to log to. This means we're also unable to obtain the `FeaturesContext`. We have to make sure we avoid logging in this scenario.

Using endpoint filters we would alternatively require:

- For invalid api keys and locked api keys, be able to write directly using IGlobalStorage. Which may require a partial rewrite.
- Investigate where in the code we call event logging, because it appears to be happening in almost every layer. So it would cause a maintenance mess.


abergs:
Otherwise I thinks this looks good!

src/AdminConsole/Middleware/EventLogStorageCommitMiddleware.cs
```diff
@@ -2,14 +2,20 @@
...
-public class EventLogStorageCommitMiddleware
+public partial class EventLogStorageCommitMiddleware
 {
     private readonly RequestDelegate _next;
 
     public EventLogStorageCommitMiddleware(RequestDelegate next) => _next = next;
 
     public async Task InvokeAsync(HttpContext context, IEventLogger eventLogger)
     {
+        if (context.GetEndpoint() == null)
```
Tyrrrz, line 13: A comment explaining why (what you wrote in the OP) would be nice

src/Api/Middleware/EventLogContextMiddleware.cs
```diff
@@ -13,6 +13,12 @@ public class EventLogContextMiddleware
 
     public async Task InvokeAsync(HttpContext context, IEventLogContext eventLogContext, IFeatureContextProvider featureContextProvider)
     {
+        if (context.GetEndpoint() == null)
```
Tyrrrz, line 16: Same here

src/Api/Middleware/EventLogStorageCommitMiddleware.cs
```diff
@@ -20,6 +20,12 @@ public EventLogStorageCommitMiddleware(RequestDelegate next, ILogger<EventLogSto
     /// <param name="provider">Used to get the `IEventLogger` instance based on the current state of the request.</param>
     public async Task InvokeAsync(HttpContext context, IServiceProvider provider)
     {
+        if (context.GetEndpoint() == null)
```
Tyrrrz, line 23: Same here

src/Service/Features/FeatureContextProvider.cs
```diff
@@ -22,17 +22,25 @@ public class FeatureContextProvider : IFeatureContextProvider
...
+                {
+                    appId = ApiKeyUtils.GetAppId(httpContext.Request.Headers["ApiSecret"].ToString());
+                }
+                else if (httpContext.Request.RouteValues.ContainsKey("appId"))
+                {
+                    appId = httpContext.Request.RouteValues["appId"].ToString();
+                }
             }
-            if (httpContext.Request.Headers.ContainsKey("ApiSecret"))
+            catch (ArgumentException)
```
Tyrrrz, line 41: Please add a comment when this exception can be thrown (i.e. why it's expected)

tests/Api.IntegrationTests/Middleware/AuthorizationIntegrationTests.cs
```diff
@@ -0,0 +1,84 @@
...
+    private readonly HttpClient _client;
+
+    public AuthorizationIntegrationTests(ITestOutputHelper testOutput, PasswordlessApiFactory apiFactory)
+    {
+        apiFactory.TestOutput = testOutput;
+        _client = apiFactory.CreateClient();
+    }
+
+    [Fact]
+    public async Task ExistentEndpointProtectedByApiKey_Returns_Unauthorized_WhenBadApiKeyIsProvided()
```
Tyrrrz, line 22: To stay consistent with our naming schema, maybe something like this (same for others):

```suggestion
    public async Task I_cannot_access_an_existing_endpoint_with_an_invalid_API_key()
```

jonashendrickx: Seems less clear to me.

Tyrrrz (approved):

PR #356, opened 2024-01-15
Add details to the error message when API key has the wrong scope.

jonashendrickx:
```json
{
    "type": "https://docs.passwordless.dev/guide/errors.html#forbidden",
    "title": "You are forbidden from viewing this resource.",
    "status": 403,
    "detail": "You are unable to view this resource because you do not have the required permissions. Required scopes: register"
}
```

Dependencies:
- [PR: Documentation](https://github.com/bitwarden/passwordless-docs/pull/82)


src/Api/Authorization/HeaderHandler.cs
```diff
@@ -89,7 +91,28 @@ protected override async Task HandleForbiddenAsync(AuthenticationProperties prop
...
+        var policies = endpoint
+            .Metadata
+            .Where(x => x is AuthorizeAttribute)
+            .Cast<AuthorizeAttribute>()
+            .ToList();
+
+        if (policies.Count > 0)
+        {
+            context.ProblemDetails.Detail =
+                $"You are unable to view this resource because you do not have the required permissions. Required scopes: {string.Join(", ", policies.Select(x => x.Policy))}";
```
Tyrrrz, line 112: Nit: maybe `access` is a better verb, considering it could be a `POST` request?

```suggestion
                $"You are unable to access this resource because you do not have the required permissions. Required scopes: {string.Join(", ", policies.Select(x => x.Policy))}";
```

src/Api/Authorization/HeaderHandler.cs
```diff
@@ -89,7 +91,28 @@ protected override async Task HandleForbiddenAsync(AuthenticationProperties prop
...
             },
-        });
+        };
+
+        var endpoint = Context.GetEndpoint();
+        if (endpoint == null)
+        {
+            await _problemDetailsService.WriteAsync(context);
+            return;
+        }
```
Tyrrrz, line 101: I would suggest un-inverting this statement (i.e. remove `return` and instead do `if (endpoint != null)`), so that `await _problemDetailsService.WriteAsync(context);` only appears once.

jrmccannon (approved):
After word change, looks good

abergs (approved):

PR #359, opened 2024-01-15
Add URL-encoding to `PasswordlessManagementClient`

Tyrrrz:
PAS-314

Also renamed some methods to have the `Async` suffix for consistency with other methods, added `using` to make proper use of the `IDisposable` pattern, and some other minor corrections.

### DIFF
--- a/src/AdminConsole/Services/PasswordlessManagement/PasswordlessManagementClient.cs
+++ b/src/AdminConsole/Services/PasswordlessManagement/PasswordlessManagementClient.cs
@@ -39,8 +39,6 @@ public class PasswordlessManagementClient(HttpClient http) : IPasswordlessManage
     public async Task<bool> DeleteApplicationAsync(string appId)
     {
         using var response = await http.DeleteAsync($"admin/apps/{Uri.EscapeDataString(appId)}");
-        // response.EnsureSuccessStatusCode(); ???
-
         return response.IsSuccessStatusCode;
     }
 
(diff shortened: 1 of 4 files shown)


src/AdminConsole/Services/PasswordlessManagement/PasswordlessManagementClient.cs
```diff
@@ -2,126 +2,160 @@
...
         return await response.Content.ReadFromJsonAsync<ICollection<string>>();
     }
 
-    public async Task<bool> DeleteApplicationAsync(string application)
+    public async Task<bool> DeleteApplicationAsync(string appId)
     {
-        var res = await _client.DeleteAsync($"admin/apps/{application}");
-        return res.IsSuccessStatusCode;
+        using var response = await http.DeleteAsync($"admin/apps/{Uri.EscapeDataString(appId)}");
+        // response.EnsureSuccessStatusCode(); ???
```
Tyrrrz, line 42: Why are we not checking the status code here? Is it intentional or an oversight?

jrmccannon (approved):

abergs (approved):

jonashendrickx (commented):
It probably needs to be sanitized server-side before it hits the PasswordlessManagementClient. With input validation in the PasswordlessManagementClient. This is likely just a placebo.

abergs:
@jonashendrickx I'm not against adding more validation and return any helpful message back, but I think this is the sane default behaviour.

PR #360, opened 2024-01-15
An atlernative for unified app settings endpoint.

jrmccannon:


src/AdminConsole/Services/PasswordlessManagement/IPasswordlessManagementClient.cs
```diff
@@ -22,4 +22,5 @@ public interface IPasswordlessManagementClient
     Task<GetAppIdAvailabilityResponse> IsApplicationIdAvailableAsync(GetAppIdAvailabilityRequest request);
     Task EnableMagicLinksAsync(string appId, string performedBy);
     Task DisableMagicLinksAsync(string appId, string performedBy);
```
Tyrrrz, line 24: Would these be removed?

src/AdminConsole/Services/PasswordlessManagement/PasswordlessManagementClient.cs
```diff
@@ -175,4 +175,9 @@ public async Task DisableMagicLinksAsync(string appId, string performedBy)
         response.EnsureSuccessStatusCode();
     }
 
+    public async Task SetAppSettingsAsync(string applicationId, SetAppSettingsRequest appSettings)
+    {
+        var response = await http.PostAsJsonAsync($"admin/apps/{applicationId}/settings", appSettings);
```
Tyrrrz:
```suggestion
        using var response = await http.PostAsJsonAsync($"admin/apps/{applicationId}/settings", appSettings);
```

Tyrrrz:
```suggestion
        var response = await http.PostAsJsonAsync($"admin/apps/{Uri.EncodeDataString(applicationId)}/settings", appSettings);
```

Tyrrrz (approved):

src/AdminConsole/Services/PasswordlessManagement/IPasswordlessManagementClient.cs
jrmccannon: probably. unless we think its useful to keep them around

abergs (commented):
I really like that the API becomes "PATCH"-like and accepts partial settings to change. 

I'm also in the "Keep things simple"-camp, preferring we keep it POST instead of going all in with HTTP verbs (since most developers don't know/care about the semantics), at least convinced otherwise.

There were some cleanup with the `switch`-statements, but other than that: Great work

src/AdminConsole/Services/PasswordlessManagement/IPasswordlessManagementClient.cs
abergs: I vote remove

jrmccannon:
I cleaned up a lot of stuff. I consolidated all the settings endpoints into the `app/features` SetFeatures endpoint. I refactored the SetFeaturesRequest to behave like a "patched" object. I added some documentation to hopefully make things a little more clear and refactored tests to reflect the changes.

Tyrrrz (approved):

PR #364, opened 2024-01-18
Normalize nullability in the database schema

Tyrrrz:
#340 fixed nullability annotations in database models, but didn't include migrations to normalize them in the database schema. This PR fixes that.


abergs (approved):
This looks like a safe migration to me

PR #365, opened 2024-01-18
Fix pin reference for HTTP request action

Tyrrrz:
Renovate relies on these comments to match hashes to tags, but this action doesn't prepend tags with `v` like the other actions.


abergs:
Ok, is there any impact of this change?

Tyrrrz:
> Ok, is there any impact of this change?

Renovate will work for this dependency.

jonashendrickx (approved):

PR #373, opened 2024-01-23
Ignore `Datadog.Trace` in Renovate updates

Tyrrrz:
This package has to be in-sync with the version of the agent we use in Azure, so we need to update it manually.

Docs reference: https://docs.renovatebot.com/configuration-options/#ignoredeps

Related: https://github.com/bitwarden/passwordless-server/pull/369

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -36,5 +36,6 @@
       "matchManagers": ["nuget"],
       "matchUpdateTypes": ["minor", "patch"]
     }
-  ]
+  ],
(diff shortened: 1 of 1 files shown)


jonashendrickx:
We should probably still be aware when this library requires updating and schedule the update for the next release.

abergs:
> We should probably still be aware when this library requires updating and schedule the update for the next release.

abergs (approved):

PR #374, opened 2024-01-23
Using existing record to cancel invite

jrmccannon:
Added some error messaging in the event a non-existent hash is passed in. Also, using the existing invite to cancel instead of passing the hashedcode through.

### DIFF
--- a/src/AdminConsole/Pages/Organization/Admins.cshtml.cs
+++ b/src/AdminConsole/Pages/Organization/Admins.cshtml.cs
@@ -133,22 +133,25 @@ public class Admins : PageModel
 
     public async Task<IActionResult> OnPostCancel(string hashedCode)
     {
-        Invites = await _invitationService.GetInvitesAsync(User.GetOrgId().Value);
+        Invites = await _invitationService.GetInvitesAsync(User.GetOrgId()!.Value);
 
         var inviteToCancel = Invites.FirstOrDefault(x => x.HashedCode == hashedCode);
 
-        await _invitationService.CancelInviteAsync(hashedCode);
+        if (inviteToCancel is null)
+        {
+            ModelState.AddModelError("error", "Could not cancel because invite could not be found.");
+            return await OnGet();
+        }
+
+        await _invitationService.CancelInviteAsync(inviteToCancel.HashedCode);
 
         var performedBy = await _dataService.GetUserAsync();
 
-        if (inviteToCancel is not null)
-        {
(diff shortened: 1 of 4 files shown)


src/AdminConsole/Pages/Organization/Admins.cshtml.cs
```diff
@@ -133,22 +133,25 @@ public async Task<IActionResult> OnPostInvite(InviteForm form)
...
     {
-        Invites = await _invitationService.GetInvitesAsync(User.GetOrgId().Value);
+        Invites = await _invitationService.GetInvitesAsync(User.GetOrgId()!.Value);
 
         var inviteToCancel = Invites.FirstOrDefault(x => x.HashedCode == hashedCode);
 
-        await _invitationService.CancelInviteAsync(hashedCode);
+        if (inviteToCancel is null)
+        {
+            ModelState.AddModelError("error", "Could not cancel because invite could not be found.");
```
Tyrrrz, line 142:
```suggestion
            ModelState.AddModelError("error", "Failed to find an invite to cancel.");
```

Tyrrrz (dismissed):

src/AdminConsole/Pages/Organization/Admins.cshtml.cs
```diff
@@ -133,22 +133,25 @@ public async Task<IActionResult> OnPostInvite(InviteForm form)
 
     public async Task<IActionResult> OnPostCancel(string hashedCode)
     {
-        Invites = await _invitationService.GetInvitesAsync(User.GetOrgId().Value);
+        Invites = await _invitationService.GetInvitesAsync(User.GetOrgId()!.Value);
```
jonashendrickx, line 136: Refactor this possibly so it doesn't require the organisation id to be passed. It's a good opportunity to touch this now.

abergs: @jonashendrickx If I reccall correctly, there's reason why we pass it down instead of through injection because the invitationService doesn't always run in a org-context.

abergs (approved):

Tyrrrz (approved):

src/AdminConsole/Pages/Organization/Admins.cshtml.cs
jrmccannon: I'm going to mark this as resolved and we can tackle it later if necessary.

PR #376, opened 2024-01-26
Adding support for multiple Postmark streams

jrmccannon:
This is a rough pass at adding the ability to add mutliple Postmark message streams. If message streams are not defined in the configuration, it will continue as normal with the current configuration.

```
  "Mail": {
    "Postmark": {
      "ApiKey": "original-key",
      "From": "do-not-reply@passwordless.com"
    }
  }
```

If message streams are provided, it will add clients specific to those streams and allow for those clients to be used based on the `MessageType` property of `MailMessage`.

```
  "Mail": {
    "Postmark": {
      "ApiKey": "original-key",
      "From": "do-not-reply@passwordless.com",
      "MessageStreams": [
        {
          "Name": "magic-links",
          "ApiKey": "something_wild",
          "From": "do-not-reply@maila.passwordless.com"
        }
      ]
    }
  }
```

The configuraiton is a bit messy and the Postmark Client factory needs some work, but this is just a WIP to determine if this is the direction we would like to go.


src/Api/Endpoints/Magic.cs
```diff
@@ -26,7 +25,7 @@ MagicLinkService magicLinkService
                 return NoContent();
             })
             .WithParameterValidation()
-            .RequireAuthorization(SecretKeyScopes.TokenVerify)
+            .RequireSecretKey()
```
abergs, line 28: looks good

src/Common/Services/Mail/PostmarkMailProvider.cs
```diff
@@ -2,33 +2,24 @@
 
 namespace Passwordless.Common.Services.Mail;
 
-public class PostmarkMailProvider : IMailProvider
+public class PostmarkMailProvider(PostmarkMailClientFactory factory) : IMailProvider
 {
```
abergs, line 6: suggestion: I'm wondering if we could remove the factory stuff and but the logic in the constructor here? Seems like we could get away with something simple and more direct:

// setup the default client
_defaultClient = defaultClient

foreach stream in Config.messageStreams
  _streamClients.add(stream.name, streamClient


and then in Send you do a streamclient.FirstOrDefault() ?? _defaultClient;

src/Common/Services/Mail/PostmarkMailProviderConfiguration.cs
```diff
@@ -0,0 +1,14 @@
...
+
+public class PostmarkMailProviderConfiguration
+{
+    public PostmarkClientConfiguration DefaultConfiguration { get; init; }
+    public List<PostmarkClientConfiguration>? MessageStreams { get; init; } = [];
+}
+
+public class PostmarkClientConfiguration
+{
+    public string Name { get; init; }
```
abergs, line 11: Should this be a property or could it instead be the key in the dictionary?

src/Common/Services/Mail/PostmarkMailProviderConfiguration.cs
```diff
@@ -0,0 +1,14 @@
+namespace Passwordless.Common.Services.Mail;
+
+public class PostmarkMailProviderConfiguration
+{
+    public PostmarkClientConfiguration DefaultConfiguration { get; init; }
+    public List<PostmarkClientConfiguration>? MessageStreams { get; init; } = [];
```
abergs, line 6: suggestion: change to dictionary?

Tyrrrz:
Why not have the following structure instead?

```json
  "Mail": {
    "Postmark": {
      "MessageStreams": [
        {
          "ApiKey": "original-key",
          "From": "do-not-reply@passwordless.com"
        },
        {
          "Name": "magic-links",
          "ApiKey": "something_wild",
          "From": "do-not-reply@maila.passwordless.com"
        }
      ]
    }
  }
```

Since the structure of the underlying objects seems the same

abergs (changes requested):

jrmccannon:
> Why not have the following structure instead?
> 
> ```json
>   "Mail": {
>     "Postmark": {
>       "MessageStreams": [
>         {
>           "ApiKey": "original-key",
>           "From": "do-not-reply@passwordless.com"
>         },
>         {
>           "Name": "magic-links",
>           "ApiKey": "something_wild",
>           "From": "do-not-reply@maila.passwordless.com"
>         }
>       ]
>     }
>   }
> ```
> 
> Since the structure of the underlying objects seems the same

This would break the current configuration. We wanted this to be backward compatible with what existed out there already.

src/Common/Services/Mail/PostmarkMailProvider.cs
jrmccannon:
```
public class PostmarkMailProvider(PostmarkMailProviderConfiguration configuration) : IMailProvider
{
    private readonly Dictionary<string, ConfiguredPostmarkClient> _clients = configuration.MessageStreams?
        .ToDictionary(x => x.Name, GetConfiguredClient) ?? new();

    private readonly ConfiguredPostmarkClient _defaultClient = GetConfiguredClient(configuration.DefaultConfiguration);
    
    public async Task SendAsync(MailMessage message)
    {
        var configuredClient = _clients.GetValueOrDefault(message.MessageType) ?? _defaultClient;
        
        PostmarkMessage pm = new PostmarkMessage
        {
            To = string.Join(',', message.To),
            From = configuredClient.From?.ToString() ?? message.From,
            Subject = message.Subject,
            TextBody = message.TextBody,
            HtmlBody = message.HtmlBody,
            Tag = message.Tag,
            Bcc = message.Bcc.Count != 0 ? string.Join(',', message.Bcc) : null,
            MessageStream = configuredClient == _defaultClient ? null : message.MessageType
        };

        IEnumerable<PostmarkResponse>? res = await configuredClient.Client.SendMessagesAsync(pm);
    }
    
    private static ConfiguredPostmarkClient GetConfiguredClient(PostmarkClientConfiguration config) =>
        new()
        {
            Client = new PostmarkClient(config.ApiKey),
            From = new MailAddress(config.From),
            Name = config.Name
        };
}
```

Here's what PostmarkMailProvider becomes.

Pros:
- Reduces everything to 1 file.

Cons:
- all clients are instantiated when the provider is created. (not a big deal now, but could be if we decide to create more of them)
- allows for a misconfiguration to go unnoticed (if a message type gets set to something that isn't configured, it will use the default and we would only notice by checking the from after the email has been sent out.

jrmccannon: We could go a bit further and move the mapping from the config file straight to the strongly typed version containing the postmark client in the DI setup code as well and just inject that object.

That way we would already have the dictionary set up and wouldn't need to do it here.

abergs: I really like the first approach, keeping it in the PostmarkMailProvider.

> Allows for a misconfiguration to go unnoticed (if a message type gets set to something that isn't configured, it will use the default and we would only notice by checking the from after the email has been sent out.

Could we add warning logging for this?

jrmccannon: Went with option 1 and added a Warning log to the provider in the event that the message type is provided for a message stream we don't currently have a config for.

abergs (approved):

PR #381, opened 2024-02-01
Fixed Attestation link size to match other "buttons"

jrmccannon:
Added `inline-block` to `btn` so that buttons have the same uniform size.

Before:
image

After:
image

### DIFF
--- a/src/AdminConsole/Pages/App/Settings/Settings.cshtml
+++ b/src/AdminConsole/Pages/App/Settings/Settings.cshtml
@@ -89,7 +89,7 @@
     <panel header="Attestation">
         <p>When attestation is enabled, you can choose to limit the authenticators used with your application.</p>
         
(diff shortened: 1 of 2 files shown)


jonashendrickx (dismissed):

src/AdminConsole/Pages/App/Settings/Settings.cshtml
```diff
@@ -89,7 +89,7 @@
     <panel header="Attestation">
         <p>When attestation is enabled, you can choose to limit the authenticators used with your application.</p>
         
-        <div>
+        <div class="flex">
```
abergs, line 92: I think we'd rather change the implementation of `.btn-primary` to use `display:inline-block`

abergs (changes requested):

abergs (approved):

PR #382, opened 2024-02-02
Merge info alert boxes into one.

jonashendrickx:
image


abergs:
I think these are suitable to be paragraphs under the title instead, so we wouldn't need the visual attention of the alert boxes. Don't you think?

jonashendrickx:
> I think these are suitable to be paragraphs under the title instead, so we wouldn't need the visual attention of the alert boxes. Don't you think?

Done @abergs.

abergs (approved):

PR #384, opened 2024-02-02
Add security headers

jonashendrickx:


src/AdminConsole/Program.cs
```diff
@@ -113,6 +113,7 @@ void RunTheApp()
     services.ConfigureApplicationCookie(o =>
     {
         o.Cookie.Name = "AdminConsoleSignIn";
+        o.Cookie.SameSite = SameSiteMode.Strict;
```
abergs, line 116: I think this is good to harden the security. We already do CSRF protection using the antiforgery-token, but having multiple levels of protection is good.

Have you tested the behaviour of the application when clicking links to the admin console? There are many places where we link or could be linking to our app (Email with magic link, from documentation, notification emails).

Are those links honoured? Or will the user be sent to a unauthorised-page, and then redirect to the index page?

CleanShot 2024-02-02 at 13 01 36

jonashendrickx: @jrmccannon or @Tyrrrz Do either of you guys want to help me test this on Monday with magic links? I can test the other scenarios.

jrmccannon: Running locally, magic links worked fine for logging into AdminConsole.

jrmccannon (approved):

src/AdminConsole/Program.cs
abergs: @jrmccannon You tested a magic link from a file open in vscode/ide, not from another website, like gmail.com -- right?

PR #388, opened 2024-02-05
Rename route parameter 'app' to 'appId'

jonashendrickx:


jrmccannon (approved):
Minor suggestion. Otherwise looks good.

PR #390, opened 2024-02-05
Active users count charts

jonashendrickx:
image

- Contains fixes for the abstract background service to calculate the initial delay correctly. With tests to back up the changes for the specific scenario's I encountered.
- BUnit testing the complete components' rendered markups. Validates that also the formatting happens correctly.
- Adding legend to charts
- Adding more colors when displaying more than one data series.


abergs:
@jonashendrickx For the Active Users-metrics, is it possible to get actual numbers too, not just the %?

I'm not 100% sure what we want here, but I think we can remove the time-series for those numbers (for now) and display an actual number + maybe the %? Not sure about the graphics, but something from https://tailwindui.com/components/application-ui/data-display/stats would be neat, maybe a mix of these?
CleanShot 2024-02-07 at 16 41 24

CleanShot 2024-02-07 at 16 41 46

src/AdminConsole/Services/ScopedPasswordlessClient.cs
```diff
@@ -89,6 +90,25 @@ public async Task<IEnumerable<PeriodicCredentialReportResponse>> GetPeriodicCred
...
+        {
+            queryBuilder.Add("from", request.From.Value.ToString("yyyy-MM-dd"));
+        }
+        if (request.To.HasValue)
+        {
+            queryBuilder.Add("to", request.To.Value.ToString("yyyy-MM-dd"));
+        }
+
+        var q = queryBuilder.ToQueryString();
+        var response = await _client.GetAsync($"/apps/{_currentContext.AppId}/reporting/active-users/periodic{q}");
```
abergs, line 106: issue, security: UrlEncode

abergs (commented):
shape, security: All Good

@Tyrrrz @jrmccannon Can you do a technical rev?

src/Service/Storage/Ef/EfGlobalGlobalStorage.cs
```diff
@@ -50,4 +50,44 @@ public async Task<int> UpdatePeriodicCredentialReportsAsync()
...
+
+                    TotalUsersCount = credentials
+                        .Select(x => x.UserId)
+                        .Distinct()
+                        .Count(),
+
+                    CreatedAt = DateOnly.FromDateTime(_timeProvider.GetUtcNow().UtcDateTime)
+                });
+
+        await _db.PeriodicActiveUserReports.AddRangeAsync(result);
```
jrmccannon, line 89: Apparently the `Async` version of this method is only to allow special value generators to access the db. Can probably switch it to the sync version.

src/Service/ReportingService.cs
```diff
@@ -31,4 +31,19 @@ public async Task<IEnumerable<PeriodicCredentialReportResponse>> GetPeriodicCred
...
+        return storage.UpdatePeriodicActiveUserReportsAsync();
+    }
+
+    public async Task<IEnumerable<PeriodicActiveUserReportResponse>> GetPeriodicActiveUserReportsAsync(PeriodicActiveUserReportRequest parameters)
+    {
+        var storage = _tenantStorageFactory.Create();
+        var entities = await storage.GetPeriodicActiveUserReportsAsync(parameters.From, parameters.To);
+        return entities
+            .Select(x => new PeriodicActiveUserReportResponse(x.CreatedAt, x.DailyActiveUsersCount, x.WeeklyActiveUsersCount, x.TotalUsersCount))
+            .ToList();
```
jrmccannon, line 47: Does the collection have to be enumerated at this point? it is already enumerated in the db so I think  this `ToList()` is unnecessary, but I could be wrong.

src/Api/Reporting/Background/ActiveUserReportingBackgroundService.cs
```diff
@@ -0,0 +1,39 @@
...
+
+public sealed class ActiveUserReportingBackgroundService : BasePeriodicBackgroundService
+{
+    private readonly IServiceProvider _serviceProvider;
+
+    public ActiveUserReportingBackgroundService(
+        IServiceProvider serviceProvider,
+        TimeProvider timeProvider,
+        ILogger<ActiveUserReportingBackgroundService> logger) : base(
+        new TimeSpan(22, 0, 0),
```
jrmccannon, line 15: Nit: I wish this was indented one more level since its part of the base class and not the current class.

src/Api/Reporting/Background/UsersCountReportingBackgroundService.cs
```diff
@@ -7,22 +7,22 @@ namespace Passwordless.Api.Reporting.Background;
 public sealed class PeriodicCredentialReportsBackgroundService : BasePeriodicBackgroundService
 {
     private readonly IServiceProvider _serviceProvider;
-    private readonly ILogger<PeriodicCredentialReportsBackgroundService> _logger;
 
     public PeriodicCredentialReportsBackgroundService(
         IServiceProvider serviceProvider,
         TimeProvider timeProvider,
         ILogger<PeriodicCredentialReportsBackgroundService> logger) : base(
         new TimeSpan(22, 0, 0),
```
jrmccannon, line 15: Nit: Same request for indentation here.

jrmccannon (approved):
Just left a few nit comments. Non-blocking. Change if you feel the need.

jrmccannon (approved):

PR #392, opened 2024-02-06
Add quotas and rate limits to magic link emails

Tyrrrz:
.


src/Service/MagicLinks/MagicLinkService.cs
```diff
@@ -1,27 +1,103 @@
...
+            // App created <3 days ago
+            < 3 when isFreeAccount => (int)(0.5 * maxFreeMinutelyRateLimit),
+            < 3 => (int)(0.5 * maxProMinutelyRateLimit),
+            // App created <30 days ago
+            < 30 when isFreeAccount => (int)(0.75 * maxFreeMinutelyRateLimit),
+            < 30 => (int)(0.75 * maxProMinutelyRateLimit),
+            // App created >30 days ago
+            _ when isFreeAccount => maxFreeMinutelyRateLimit,
+            _ => maxProMinutelyRateLimit
+        };
```
Tyrrrz, line 88: @abergs does this look about right? The numbers in the original pseudocode didn't line up, so I took the liberty to tweak them myself

src/Service/Storage/Ef/EfTenantStorage.cs
```diff
@@ -285,6 +285,31 @@ public Task RemoveAuthenticatorsAsync(IEnumerable<Guid> aaGuids)
...
+        await db.SaveChangesAsync();
+
+        return email;
+    }
+
+    public async Task<int> GetDispatchedEmailCountAsync(TimeSpan duration)
+    {
+        var from = _timeProvider.GetUtcNow().UtcDateTime - duration;
+        return await db.DispatchedEmails
+            .Where(x => x.CreatedAt >= from)
```
jrmccannon, line 309: This query could get quite expensive quickly. MSSQL is not so good at creating a good query plan over a time range.

We could consider a view instead that would only be on SQL and it would be better and determining the plan. We could then query one record by application containing it's current minute total and it's current monthly total. Maybe an early optimization but I just get worried about querying against time spans. Especially since I don't see a delete after an amount of time.

src/Service/MagicLinks/MagicLinkService.cs
jonashendrickx: Perhaps you can go with a strategy pattern to differentiate between free and paid and keep the rest.

src/Service/MagicLinks/MagicLinkService.cs
```diff
@@ -1,27 +1,98 @@
...
-
-    public MagicLinkService(IFido2Service fido2Service, IMailProvider mailProvider)
+    private async Task EnforceLimitsAsync(MagicLinkDTO dto)
     {
-        _fido2Service = fido2Service;
-        _mailProvider = mailProvider;
+        var now = timeProvider.GetUtcNow();
+        var account = await tenantStorage.GetAccountInformation();
+        var accountAge = (now - account.CreatedAt).Duration();
+        var isFreeAccount = account.SubscriptionTier == "Free";
```
jonashendrickx, line 20: @Tyrrrz & @abergs The 'SubscriptionTier' in the `Api` is no longer being updated. The `Api` has no concept of what plan a customer is on. The architectural proposal for self-hosting licensing was meant to change all this.

- [PR](https://github.com/bitwarden/passwordless-server/pull/320)

jonashendrickx: Upon longer inspection, you'll need to insert the column `MaxFreeMonthlyLimit` in the `AppFeatures` table.
Then update the appsettings.json to set the values for each plan accordingly.
Then make sure the mapping everywhere happens correctly when an app is created or on a plan change.

Use a good prefix so we know it's used for magic links.

jonashendrickx (changes requested):

src/Service/MagicLinks/MagicLinkService.cs
jonashendrickx: Get rid of this complex switch case, and use AppFeatures, see below

Tyrrrz:
After some brainstorming with @jrmccannon, I have decided to not introduce any caching logic for now. The reasons are:

1. Ideally we would need not a cache but a replica that would keep the list of the last 30 days worth of sent emails, and update them alongside the database. That way we can reduce all DB queries for rate limiting to only the initial one.
2. Doing the above is quite tricky as it involves sharing state between multiple scoped services, grouped by tenants. Shared state and mutations can be difficult.
3. Whatever caching solution we go for, we have to consider scaling issues. If we want to keep track of 30 days worth of emails for each tenant, it will get costly in terms of memory.
4. EF Core already has some caching mechanisms built-in, so it's possible/likely that it would optimize the query on its own. If that's the case, we would just be solving a non-existent problem.

So in the meantime we can observe the performance impact through DD and come up with a new plan if the need arises.

src/Service/Storage/Ef/EfTenantStorage.cs
Tyrrrz: Added a top-level comment based on our discussion

src/Service/Models/AppFeature.cs
```diff
@@ -9,6 +9,18 @@ public class AppFeature : PerTenant
     /// </summary>
     public int EventLoggingRetentionPeriod { get; set; }
 
+    /// <summary>
+    /// Maximum monthly limit for magic link emails sent by this application.
+    /// The actual limit may be lower, depending on the age of the application.
+    /// </summary>
+    public int MaxMagicLinkEmailMonthlyLimit { get; set; }
```
jrmccannon, line 16: Rename to `MagicLinkEmailMaxMonthlyLimit`

src/Service/Models/AppFeature.cs
```diff
@@ -9,6 +9,18 @@ public class AppFeature : PerTenant
...
+
+    /// <summary>
+    /// By-minute rate limit for magic link emails sent by this application.
+    /// The actual limit may be lower, depending on the age of the application.
+    /// </summary>
+    public int MaxMagicLinkEmailMinutelyLimit { get; set; }
```
jrmccannon, line 22: Rename to `MagicLinkEmailMaxMinutelyLimit`

src/Service/MagicLinks/MagicLinkService.cs
Tyrrrz: Simplified the switch expression

src/Service/MagicLinks/MagicLinkService.cs
Tyrrrz: Done

src/Service/MagicLinks/MagicLinkService.cs
```diff
@@ -1,27 +1,89 @@
...
+            < 1 => 0.2 * maxMinutelyLimit,
+            // App created <3 days ago
+            < 3 => 0.5 * maxMinutelyLimit,
+            // App created <30 days ago
+            < 30 => 0.75 * maxMinutelyLimit,
+            // App created >30 days ago
+            _ => maxMinutelyLimit
+        };
+
+        if (await tenantStorage.GetDispatchedEmailCountAsync(TimeSpan.FromDays(30)) >=
```
abergs, line 68: suggestion: `TimeSpan.FromDays(30)` should maybe be something like `TimeSpan.FromMinutes(5)`

If this is the "tempo" regulator (as in: you can't send all your emails in a 5 minute window), I'm not sure you want fetch all the 30 days?

src/Service/Models/DispatchedEmail.cs
```diff
@@ -0,0 +1,14 @@
...
+
+public class DispatchedEmail : PerTenant
+{
+    public required Guid Id { get; set; }
+
+    public required DateTimeOffset CreatedAt { get; set; }
+
+    public required string UserId { get; set; }
+
+    public required string EmailAddress { get; set; }
```
abergs, line 11: thought: We should make UserId and EmailAdress optional in the app, make it opt-out.

issue: We probably want to track the outgoing domain/url for monitoring missbehaviour.

abergs:
@Tyrrrz **I support shipping without a memorycache** - I'm just adding context to any future discussions:

Why would we need a replica?
I think maybe a counter with non-sliding expiration of <24h would do?

Scenario:

Tenant is about to send email #31:
* Empty cache
* Reads current consumption from db, 30 emails sent
* Check is OK, sends email
* Records the email to db and cache _cache.Set("emails_sent_this_month", current + 1, expire: TimeSpan.FromHours(24))

enant is about to send email #31:
* Cache value = 31
* Reads current consumption from db, 31 emails sent
* Check is OK, sends email
* Records the email to db and cache _cache.Set("emails_sent_this_month", current + 1) # don't restart the cache expiration

If we don't distribute the cache across all servers, it will be out of sync for the duration of the lifetime. Which is not a critical bug -- but a trade-off. A shorter lifetime would help the individual caches stay closer to reality.

src/AdminConsole/Billing/Configuration/FeaturesOptions.cs
```diff
@@ -6,6 +6,18 @@ public class FeaturesOptions
...
+    /// Maximum monthly limit for magic link emails sent by this application.
+    /// The actual limit may be lower, depending on the age of the application.
+    /// </summary>
+    public int MagicLinkEmailMaxMonthlyLimit { get; set; }
+
+    /// <summary>
+    /// By-minute rate limit for magic link emails sent by this application.
+    /// The actual limit may be lower, depending on the age of the application.
+    /// </summary>
+    public int MagicLinkEmailMaxMinutelyLimit { get; set; }
```
abergs, line 19: I'm wondering if we should rename this to be called Rate instead of Minutely?

src/Service/MagicLinks/MagicLinkService.cs
Tyrrrz: The 30 days is required for monthly quota

src/Service/Models/DispatchedEmail.cs
Tyrrrz: Technically, I only need the date here. I thought the additional fields would be useful if we need to identify the email. Should I remove them?

src/AdminConsole/Billing/Configuration/FeaturesOptions.cs
Tyrrrz: It made sense to name those two properties similarly because they work essentially the same way, with the only difference being the time window

Tyrrrz:
> @Tyrrrz **I support shipping without a memorycache** - I'm just adding context to any future discussions:
> 
> Why would we need a replica? I think maybe a counter with non-sliding expiration of <24h would do?

A replica would allow us to keep track of sent emails without re-querying the database, for as long as the application is running. Technically, EF already has that in the form of `LocalView` (i.e. `DbSet.Local`), but I don't know what are the retention policies on that. The main challenge is that we need to keep track of when each individual email is sent.

> Scenario:
> 
> Tenant is about to send email #31:
> 
> * Empty cache
> * Reads current consumption from db, 30 emails sent
> * Check is OK, sends email
> * Records the email to db and cache _cache.Set("emails_sent_this_month", current + 1, AbsoluteExpiration: TimeSpan.FromHours(24))

This was an idea I had too. Note that the initial cache lifetime needs to be synced with the current time, ideally.

However, to replicate this approach for the minutely rate limits, we would use caches with lifetimes of less than a minute (ideally 1 second, but probably 10-30 seconds are okay too). This might invalidate the point of having those caches in the first place.

> enant is about to send email #31:
> 
> * Cache value = 31
> * Reads current consumption from db, 31 emails sent
> * Check is OK, sends email
> * Records the email to db and cache _cache.Set("emails_sent_this_month", current + 1) # don't restart the cache expiration
> 
> If we don't distribute the cache across all servers, it will be out of sync for the duration of the lifetime. Which is not a critical bug -- but a trade-off. A shorter lifetime would help the individual caches stay closer to reality.

Hmm, why would we still read from the DB in the second step here?

---

Maybe as a different approach, we can consider using @abergs's suggestion for caching quotas, but change the minutely rate limits so that they apply to the endpoint requests, rather than successfully dispatched emails. That would allow us to configure a reasonable limit (maybe even vary by tenant, since we only need to cache their `AccountInformation`), without consulting the database on each attempt to send an email.

So instead of: **5 emails sent** per minute, it would be **5 attempts to send an email** (i.e. endpoint requests) per minute

abergs:
> change the minutely rate limits so that they apply to the endpoint requests, rather than successfully dispatched emails. That would allow us to configure a reasonable limit (maybe even vary by tenant, since we only need to cache their AccountInformation), without consulting the database on each attempt to send an email.
So instead of: 5 emails sent per minute, it would be 5 attempts to send an email (i.e. endpoint requests) per minute

@Tyrrrz 100% yes. The rate-limit is just to fend off bad actors trying to send a lot of emails in a short time. I'm not even sure we need to store the configuration in a database - only reason would be to allow apps that has proven to be trustworthy to have larger boundaries (e.g. large enterprise app that has spiky traffic patterns should not be rate limited in the same way as free app)


edit:
> Hmm, why would we still read from the DB in the second step here?
Sorry, my misstake, I've edited my reply.

Tyrrrz:
Sorry for the churn in tests, had to rework the infra so that each test gets its own version of the API server and can advance the time independently.

Tyrrrz:
Current issue: the built-in rate limiter that comes with ASP.NET Core does not seem to be using `TimeProvider`, so I can't test it. Moreover, the monthly quota tests are also not accurate because they're getting 429 from the rate limiter instead of the quota (since the former kicks in first).

Not sure what to do about it now, going to sleep on it.

cc @abergs

abergs:
> Not sure what to do about it now, going to sleep on it.

I think you could just set the quota really low and send one or two emails, that should hit it?

Tyrrrz:
Okay, the rate limiter can now be disabled selectively for tests. I could not test rate limit recovery because, again, can't inject a `TimeProvider` in there. But at least the quota and quota recovery are properly tested now.

src/AdminConsole/appsettings.json
```diff
@@ -48,6 +49,7 @@
         "Features": {
           "EventLoggingIsEnabled": false,
           "EventLoggingRetentionPeriod": 0,
+          "MagicLinkEmailMonthlyQuota": 1000,
           "MaxAdmins": 1000,
           "MaxApplications": 1000
         },
```
abergs, line 55: Missing quota for enterprise, set to same as Pro.
issue: Let's lower free emails to 100.

src/Api/Email/DispatchedEmailCleanupService.cs
```diff
@@ -0,0 +1,12 @@
...
+
+namespace Passwordless.Api.Email;
+
+public class DispatchedEmailCleanupService(TimeProvider timeProvider, ITenantStorage tenantStorage)
+    : BasePeriodicBackgroundService(new TimeOnly(00, 00, 00), TimeSpan.FromDays(1), timeProvider)
+{
+    protected override async Task DoWorkAsync(CancellationToken cancellationToken) =>
+        // We only need 30 days worth of emails, but let's keep a small extra buffer just in case
+        await tenantStorage.DeleteOldDispatchedEmailsAsync(TimeSpan.FromDays(50));
+}
```
abergs, line 12: Nice.

src/Service/MagicLinks/MagicLinkService.cs
```diff
@@ -1,27 +1,84 @@
...
+                cacheEntry.SetAbsoluteExpiration(TimeSpan.FromDays(1));
+                return await tenantStorage.GetDispatchedEmailCountAsync(TimeSpan.FromDays(30));
+            }
+        );
+
+        if (emailsDispatchedIn30Days >= quota)
+        {
+            throw new ApiException(
+                "You have reached your monthly quota for magic link emails. " +
+                "Please try again later.",
```
abergs, line 68: suggestion: Should we have a error_code for this error and link to the docs so the developer can read up on when the quota reset?

src/Service/Models/DispatchedEmail.cs
abergs: @Tyrrrz We can keep them for now, but definitely add UrlTemplate here (without the tokens)

abergs (commented):
shape: solid!

technical: See comments

src/Service/MagicLinks/MagicLinkService.cs
Tyrrrz: Maybe. How do I go about adding a new error code?

src/AdminConsole/appsettings.json
Tyrrrz: Done

src/Service/Models/DispatchedEmail.cs
Tyrrrz: RIP migrations again 😂

Tyrrrz: Done

src/AdminConsole/Services/BaseBillingService.cs
```diff
@@ -54,20 +53,21 @@ protected async Task SetPlanOnApp(string app, string selectedPlan, string subscr
         {
             EventLoggingIsEnabled = plan.Features.EventLoggingIsEnabled,
             EventLoggingRetentionPeriod = plan.Features.EventLoggingRetentionPeriod,
+            MagicLinkEmailMonthlyQuota = plan.Features.MagicLinkEmailMonthlyQuota,
```
jonashendrickx, line 56: Also handle `public async Task OnSubscriptionDeletedAsync(string subscriptionId)`.

jonashendrickx (changes requested):

jonashendrickx:
Check  `Passwordless.AdminConsole.Pages.Organization` for the file `CreateApplication.cshtml.cs`. You need to map the new field there as well.

jonashendrickx:
More missing mapping in SharedManagementService:

- `Task SetFeaturesAsync(ManageFeaturesRequest features)`: If you want it to be saved upon plan change.
- `Task SetFeaturesAsync(SetFeaturesRequest features)`: Should only be mapped here if you intend for the customer to be able to change the value himself.

Tyrrrz:
> More missing mapping in SharedManagementService:
> 
> * `Task SetFeaturesAsync(ManageFeaturesRequest features)`: If you want it to be saved upon plan change.
> * `Task SetFeaturesAsync(SetFeaturesRequest features)`: Should only be mapped here if you intend for the customer to be able to change the value himself.

This is already in place. The user shouldn't be able to set their own quota.

jrmccannon (dismissed):

jonashendrickx (approved):

jonashendrickx:
@Tyrrrz I fixed the merge conflict.

jonashendrickx (dismissed):

src/Service/Storage/Ef/DbGlobalContext.cs
```diff
@@ -101,6 +102,15 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
 
+        modelBuilder.Entity<DispatchedEmail>(builder =>
+        {
+            builder.HasKey(x => x.Id);
+            builder.HasOne(x => x.Application)
+                .WithMany(x => x.DispatchedEmails)
+                .HasForeignKey(x => x.Tenant)
+                .IsRequired();
+        });
+
```
abergs, line 113: @Tyrrrz Saw something here, we should probably index on Created since that is how we will query.

abergs (commented):
technical: See comment

abergs (commented):
technical: See comment

src/Service/Storage/Ef/DbGlobalContext.cs
Tyrrrz: Done

jrmccannon (dismissed):

abergs (approved):

abergs (dismissed):

PR #396, opened 2024-02-09
BEEEP: Move middleware to maintain readability.

jonashendrickx:
I was having a hard time finding the middleware files when I needed to. This will make the code also easier to maintain in the future as it simplifies the way the dependencies are injected for the content security policy middleware.


src/AdminConsole/Middleware/SecurityHeadersMiddleware.cs
```diff
@@ -0,0 +1,19 @@
...
+    private readonly RequestDelegate _next;
+
+    public SecurityHeadersMiddleware(RequestDelegate next)
+    {
+        _next = next;
+    }
+
+    public void Invoke(HttpContext context)
+    {
+        context.Response.Headers.Append("X-Content-Type-Options", new[] { "nosniff" });
```
abergs, line 14: before-merging: Make sure you an admin can still download their API keys, this could potentially impact that.

jonashendrickx: @abergs It was just moved from UseCsp.cs

jonashendrickx:
> @jonashendrickx Do you think its valuable to keep this as two different middlewares? To me it seems like we could move csp into the SecurityHeadersMiddleware

@abergs Yes, might be a good idea

abergs (approved):

abergs (approved):
shape, security: all good

PR #401, opened 2024-02-10
Prepare to kill `EfTenantStorageFactory`

jonashendrickx:
In reality, we do not need the `EfTenantStorageFactory`, as we can have it created by the DI container. The [tenant query filter](https://github.com/bitwarden/passwordless-server/pull/401/files#diff-87552d446dc666649a8753d567aa99814b0e6b492334236e713e649008e99e4a) could be set at a later point with [TenantProvider](https://github.com/bitwarden/passwordless-server/pull/401/files#diff-e95e4f237f39bf77cd7b7f7215933fca86f2558c39ab1e3a77893f7efb8f50b1) where the getter of the property reads the accountName claim. This is basically done because when the services are created, the API key may not have been validated yet, i.e. the Authorization middleware hasn't run yet. So basically it would then crash saying you should provide a valid tenant.

When you get that sorted out, you'll stumble over the Authorization middleware, which is actually using the [EfTenantStorageFactory](https://github.com/bitwarden/passwordless-server/pull/401/files#diff-8fea102b21422a3e63bcd99db44fec073ba9bd559b534cfd3dfe1deaf6f08498) to create the [EfTenantStorage](https://github.com/bitwarden/passwordless-server/pull/401/files#diff-3f0eb63012da38eaf10463c8b9bffcaebafd1fa8db019f42c5e8f020dc9004af) by manually specifying the `AppId` or tenant from the API key. If you want to read it securely from the claim, it won't work, because at this point the claims haven't been populated yet. Either you would have to keep the EfTenantStorageFactory here and manually specify the app id, or use the EfGlobalStorage.

Once this is sorted out, you're good to go. If you then move all endpoints with the management key to use the secret key where they are app scoped, you can then securely instantiate the EfTenantStorage from the claim, and have zero risk for RPO or impersonation-like vulnerabilities where you can access someone else's app.

Refactoring further, you'd have a unified way to access the app id, and you can greatly simplify the middleware everywhere. We only leave managementkey endpoints where requests are not app-scoped or simply they serve to create an application.

We can still keep a blacklist somewhere of scopes a customer is not able to request to keep certain endpoints internal.

The EfGlobalStorageFactory has been similarly removed.

Additional fixes and simplifications were brought to the health checks.


src/Api/Helpers/AddDatabaseExtensionMethod.cs
```diff
@@ -29,6 +27,7 @@ public static IServiceCollection AddDatabase(this IServiceCollection services, I
                 // resolving config from SP to avoid capturing
                 builder.UseSqlite(sp.GetRequiredService<IConfiguration>().GetConnectionString("sqlite:api"));
             });
+            services.AddScoped<ITenantStorage, EfTenantStorage>();
```
abergs, line 30: nitpitck, duplicated: We can de-duplicate this row by moving it of the IF, e.g. together with ITenantProvider.

src/Api/Helpers/AddDatabaseExtensionMethod.cs
```diff
@@ -44,22 +43,15 @@ public static IServiceCollection AddDatabase(this IServiceCollection services, I
                 // resolving config from SP to avoid capturing
                 builder.UseSqlServer(sp.GetRequiredService<IConfiguration>().GetConnectionString("mssql:api"));
             });
+            services.AddScoped<ITenantStorage, EfTenantStorage>();
```
abergs, line 46: duped location

src/Service/Features/FeatureContextProvider.cs
```diff
@@ -47,7 +47,7 @@ public class FeatureContextProvider : IFeatureContextProvider
 
             if (string.IsNullOrWhiteSpace(appId)) return new NullFeaturesContext();
 
-            var storage = _storageFactory.Create(appId);
+            var storage = _tenantStorageFactory.Create(appId);
```
abergs, line 50: question: Does this run after Authz? Is that why we can't inject ITenantStorage directly?

src/Service/ITenantProvider.cs
```diff
@@ -5,5 +5,5 @@ namespace Passwordless.Service;
 
 public interface ITenantProvider
 {
-    public string Tenant { get; set; }
+    public string Tenant { get; }
```
abergs, line 8: Neat

src/Service/SharedManagementService.cs
```diff
@@ -98,8 +101,7 @@ public async Task<CreateAppResultDto> GenerateAccount(string appId, CreateAppDto
             throw new ApiException("accountName needs to be alphanumeric and start with a letter", 400);
         }
 
-        ITenantStorage storage = tenantFactory.Create(accountName);
-
+        var storage = tenantFactory.Create(appId);
```
abergs, line 104: This seems reasonable.

src/Service/SharedManagementService.cs
```diff
@@ -142,8 +144,8 @@ public async Task<CreateAppResultDto> GenerateAccount(string appId, CreateAppDto
     public async Task<ValidateSecretKeyDto> ValidateSecretKey(string secretKey)
     {
         var appId = GetAppId(secretKey);
-        var storage = tenantFactory.Create(appId);
 
+        var storage = _globalStorageFactory.Create();
```
abergs, line 148: non-blocking, separate-issue: This might be the subject of a separate PR and issue but: I can't see why we need a globalStorageFactory instead of injecting it?

src/Service/Storage/Ef/DbTenantFactory.cs
```diff
@@ -4,7 +4,13 @@ namespace Passwordless.Service.Storage.Ef;
...
 {
-    ITenantStorage Create();
+    /// <summary>
+    /// Use <see cref="ITenantStorage" /> instead when not calling from `ManagementKey` scope.
+    /// <see cref="ITenantStorage" /> uses the `accountName` claim in <see cref="TenantProvider" /> to determine the tenant when an authorization is successful.
+    /// </summary>
+    /// <param name="appId"></param>
+    /// <returns></returns>
+    [Obsolete("Use injected `ITenantStorage` instead when not calling from `ManagementKey` scope.")]
     ITenantStorage Create(string appId);
```
abergs, line 14: praise: Good, this will help us from making this misstake

src/Service/Storage/Ef/DbTenantFactory.cs
```diff
@@ -20,17 +26,9 @@ public EfTenantStorageFactory(IServiceProvider services, TimeProvider timeProvid
...
     }
 
-    public ITenantStorage Create()
-    {
-        var tenantProvider = _services.GetRequiredService<ITenantProvider>();
-        var context = ActivatorUtilities.CreateInstance<TContext>(_services, tenantProvider);
-
-        return new EfTenantStorage(context, _timeProvider);
-    }
-
```
abergs, line 30: praise: good removal

src/Service/Storage/Ef/EfGlobalGlobalStorage.cs
```diff
@@ -51,6 +52,13 @@ public async Task<int> UpdatePeriodicCredentialReportsAsync()
         return rows;
     }
 
+    public Task<ApiKeyDesc> GetApiKeyAsync(string apiKey)
+    {
+        var appId = ApiKeyUtils.GetAppId(apiKey);
+        var pk = apiKey.Substring(apiKey.Length - 4);
+        return _db.ApiKeys.FirstOrDefaultAsync(e => e.Id == pk && e.Tenant == appId);
+    }
+
```
abergs, line 61: question: While OK, just curious, why was this moved?

src/Service/TenantProvider.cs
```diff
@@ -1,8 +1,26 @@
...
+
+    public string Tenant
+    {
+        get
+        {
+            var httpContext = _httpContextAccessor.HttpContext;
+            var accountName = httpContext?.User.FindFirstValue("accountName");
+            return accountName ?? throw new InvalidOperationException("Tenant not found");
+        }
+    }
```
abergs, line 25: security: All good.

src/Api/HealthChecks/DatabaseHealthCheck.cs
```diff
@@ -5,21 +5,20 @@ namespace Passwordless.Api.HealthChecks;
...
         _logger = logger;
     }
 
     public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
     {
         try
         {
-            var storage = _tenantStorageFactory.Create(null);
-            await storage.GetAccountInformation();
+            await _tenantStorage.GetAccountInformation();
```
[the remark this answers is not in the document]
jonashendrickx: @abergs Fixed it. It will also validate the connection to the database.

[Source](https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-8.0#entity-framework-core-dbcontext-probe)

I don't recall the option being there in .NET 6 or .NET 7.

src/Api/Helpers/AddDatabaseExtensionMethod.cs
jonashendrickx: @abergs done

src/Service/Features/FeatureContextProvider.cs
jonashendrickx: This can even run before authentication. It's used by event logging. Could even be triggered when logging bad api keys for example. When that happens, the claims will not be populated.

Some endpoints in the admin console also call the get features endpoint which leverages the FeaturesContextProvider. The reason that was done there was that the FeaturesContextProvider could have already been instantiated in the middleware, so the `FeaturesContext` is already in memory. So it was much quicker to access the `FeaturesContextProvider`.

src/Service/ITenantProvider.cs
jonashendrickx: The `ManualTenantProvider` is using the constructor, which is used for the EfTenantStorageFactory that remains. Because the TenantProvider reads out the claims in the getter.

src/Service/Storage/Ef/EfGlobalGlobalStorage.cs
jonashendrickx: When validating the api keys and api secrets, there is no valid authorization. Claims are not populated, so we cannot use the EfTenantStorage. If we choose to keep the EfTenantFactory, we could set the appId manually with ManualTenantProvider. And then we could keep it on EfTenantStorage.

src/Service/SharedManagementService.cs
jonashendrickx: I think I did it for consistency, but I can look into it.

src/Service/Storage/Ef/EfGlobalGlobalStorage.cs
abergs: Yeah I think I'd prefer this code to be part of the TenantStorage, and use the factory to get a TenantStorage while validating them. Just makes it less of a risk that we accidentally call this method.

jonashendrickx: Ok moved.

abergs:
@bitwarden/team-passwordless-dev This looks good to me, but would probably make sense to have a second person reviewing since the change is large-ish.

jrmccannon (approved):

PR #410, opened 2024-02-13
Clean up 'Playground - Registration'

jonashendrickx:


src/AdminConsole/Pages/App/Playground/NewAccount.cshtml
```diff
@@ -14,7 +13,7 @@
 }
 
 
-<div id="xapp">
+<div>
```
jrmccannon, line 16: This outer div isn't needed anymore.

jrmccannon (approved):
It breaks my heart to remove Vue. 💔

PR #412, opened 2024-02-14
API Key Management Confirmation Dialog & Horizontal Scrolling

jonashendrickx:
https://github.com/bitwarden/passwordless-server/assets/1045327/857a970b-b4fc-4ea2-a740-c8584125b902

- There is some backwards compatibility to make the form binding happen when the parent is a Razor Page. When we use a Blazor page the block can be removed. Other than that, behaves completely as normal.
- The modal is very flexible to implement.
- Refactoring removes any code relating to API Key Management from `Settings.cshtml` and `Settings.cshtml.cs`. Isolating it completely. As our settings page keeps growing this will make it easier to maintain. It only takes 1 minute to port this to a completely new page when deemed necessary in the future. Razor Pages requires always the page handler to be implemented on the page itself rather than the partial view. Which can remove single responsibility and cause business logic to leak.
- Incorporated the new table also in the ManageAuthenticators.razor page as it was also missing horizontal scrolling.

Flow for locking/unlocking/deleting:

1. Click one of the action buttons.
2. A POST will happen to the page with the form data, causing the dialog to render. However it will be intercepted with javascript.
3. Clicking confirm will allow the form submission to continue to happen. Cancel will just hide the dialog

**Remarks**

- We should absolutely consider porting other pages to use the `Table` component, as I've had to fix too many tables in the pas 6 months to get a more uniform design and behavior. One of the present limitations of Razor Pages is that we cannot set the `RenderFragment` or `ChildContent` property for the body of the table if we wish to use the component.
- I left the content cells to use `td` elements. Too early to tell the pattern that we need. Likely we need a DateTimeCell, TextCell, ActionCell, etc to enforce a consistent layout inside.


jrmccannon (dismissed):

abergs:
I want to review this before we merge

Tyrrrz:
The warning icon seems to be misaligned with the text, or is it intentional?

image

jonashendrickx:
> The warning icon seems to be misaligned with the text, or is it intentional?
> 
> image

We are paying for these components. Maybe intentional? I only have one paragraph vs three in the sample.

[Source](https://tailwindui.com/components/application-ui/overlays/dialogs)

jonashendrickx:
> The warning icon seems to be misaligned with the text, or is it intentional?
> 
> image

Alignment is fixed, I removed the `mt-3`.

jrmccannon (approved):

PR #415, opened 2024-02-16
AppId availability check is returning a 500 in admin console.

jonashendrickx:
`GET /admin/apps/{appId}/available` returns a `409` when checking whether an `AppId` is available. However, since we're not modifying a resource, it should just return a `200` instead, with the expected body.

Also implemented the problem details handler, which I believe I did also before, but may have been removed. It will now properly catch any invalid payloads and throw a `PasswordlessApiException` instead of a `HttpRequestException`.

`AppId` with length less than 3 characters will return a `400` in problem details format.

Our observability tooling should no longer display unnecessary errors.


src/AdminConsole/Services/PasswordlessManagement/ProblemDetailsDelegatingHandler.cs
abergs, line 1: This is indeed implemented in the .net sdk, I suggest we try to re-use it: https://github.com/bitwarden/passwordless-dotnet/blob/854d3e5f4695318b8d8bff7edf4317f4a8097d3b/src/Passwordless/PasswordlessHttpHandler.cs#L30

abergs (commented):
Assuming the issue in the comment is addressed, this looks fairly straight forward(?)

I'll let someone else double check

src/AdminConsole/Services/PasswordlessManagement/ProblemDetailsDelegatingHandler.cs
jonashendrickx: @abergs It is now using `PasswordlessHttpHandler`.

src/AdminConsole/Endpoints/ApplicationEndpoints.cs
```diff
@@ -13,8 +13,18 @@ public static IEndpointRouteBuilder MapApplicationEndpoints(this IEndpointRouteB
...
+                var result = await client.IsApplicationIdAvailableAsync(request);
+                return Ok(result);
+            }
+            catch (PasswordlessApiException e)
+            {
+                return Json(
+                    e.Details,
+                    contentType: "application/problem+json",
+                    statusCode: e.Details.Status);
+            }
```
Tyrrrz, line 27: In API, we do this inside a middleware for all endpoints. Maybe we should do that here too?

src/Api/Endpoints/Apps.cs
```diff
@@ -127,6 +116,12 @@ public static void MapAccountEndpoints(this WebApplication app)
             .RequireCors("default");
     }
 
+    public static async Task<IResult> IsAppIdAvailableAsync([AsParameters] GetAppIdAvailabilityRequest payload, ISharedManagementService accountService)
```
Tyrrrz, line 119: Personally prefer the method to be inlined as it was before

src/AdminConsole/Endpoints/ApplicationEndpoints.cs
abergs: @Tyrrrz Not sure I'm thinking correctly, but: Adminconsole doesn't return Json in most cases (this being a bit special endpoint). I think(?) we have a generic exception page that gets rendered for uncaught exceptions

Tyrrrz: Ah I see. Makes sense.

src/Api/Endpoints/Apps.cs
jonashendrickx: No

abergs:
shape: I'm OK with these changes. I'll delegate other review/approval to @Tyrrrz or @jrmccannon

jrmccannon (approved):

Tyrrrz (approved):

PR #416, opened 2024-02-16
Soften the logging level when user is not found.

jonashendrickx:
Depending on the MagicLinkBuilder to throw an exception seems like a bad idea. So we check early whether the user is null, and return signin as failed accordingly.


abergs (commented):
Closing this due to efforts in #421

PR #421, opened 2024-02-16
Use Magic Links in AdminConsole

abergs:
This PR contains a draft implementation of Magic Links in the AdminConsole.

- [ ] Remove "MagicClient" when implementation is in SDK
- [ ] Test the ManuallyGenerated links
- [x] Set longer expiration on auth tokens.
- [x] Nitpick: Add a bit of random time-delay in "no user exists" to reduce race-enumeration

Side effects

* Magic Links now end up in Api/ -- do we need to update docs/vscode link for dev env?


src/AdminConsole/Pages/Initialize.cshtml.cs
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public void OnGetAsync(CreateAppResultDto keys, string managementKey, string apiUrl)
+    {
+        CreateAppResultDto = keys;
+        ManagementKey = managementKey;
+        ApiUrl = apiUrl;
+    }
+
+    public async Task OnPostAsync()
```
jonashendrickx, line 34: You might want to use `Blazor` for this, so you can use the form confirmation dialog.

Also generate a random `appid`. This creates an enumeration vulnerability.

jonashendrickx (changes requested):

src/AdminConsole/Pages/Initialize.cshtml.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        using var http = new HttpClient
+        {
+            BaseAddress = new Uri(ApiUrl),
+            DefaultRequestHeaders = { { "ManagementKey", ManagementKey } }
+        };
+
+        using var response = await http.PostAsJsonAsync("/admin/apps/AdminConsole/create", request);
+
+        var keys = await response.Content.ReadFromJsonAsync<CreateAppResultDto>();
```
jonashendrickx, line 49: Once, posted successfully, render a button `Verify configuration`, which performs a check using javascript whether you can call the 'admin' app. When successful, redirect to a page `/SetupCompleted`.

jonashendrickx (changes requested):

src/AdminConsole/Pages/Initialize.cshtml.cs
jonashendrickx: Also assign an AuthorizationHandler to the two setup pages, to ensure that they are not working once the apps are created.

src/AdminConsole/Pages/Initialize.cshtml.cs
abergs: Any specific reason to use form confirmation? I think we're good without it.

And I don't see any enumeration problem with this?

src/AdminConsole/Pages/Initialize.cshtml
```diff
@@ -0,0 +1,31 @@
...
+        <input asp-for="AdminEmail" type="email"/>
+    </label>
+    <label asp-for="ApiUrl">
+        <input asp-for="ApiUrl" type="url"/>
+    </label>
+    <button type="submit"></button>
+</form>
+
+Paste these values into where you configure your AdminConsole
+<code lang="json">
```
abergs, line 21: curious: Does this give us syntax highlighting? Haven't seen lang="json" used before

src/AdminConsole/Pages/Initialize.cshtml.cs
abergs: > Once, posted successfully, render a button Verify configuration, which performs a check using javascript whether you can call the 'admin' app

This is a neat idea, I'd suggest creating a ticket for it and not blocking this merge. 

> AuthorizationHandler

I think we could just add a `if has apikey` throw exception/return null? Seems easy enough.

src/AdminConsole/Pages/Initialize.cshtml
```diff
@@ -0,0 +1,31 @@
...
+  "Passwordless": {
+    "ApiUrl": "@Model.ApiUrl",
+    "ApiKey": "@Model.CreateAppResultDto.ApiKey1",
+    "ApiSecret": "@Model.CreateAppResultDto.ApiSecret1"
+  },
+  "PasswordlessManagement": {
+    "ApiUrl": "@Model.ApiUrl",
+    "ManagementKey": "@Model.ManagementKey"
+  }
```
abergs, line 30: We should absolutely not render the ManagementKey. And the ApiUrl would already have needed to be configured for this to work (at least the PasswordlessManagement" one, so that whole section can go.

src/AdminConsole/Pages/Initialize.cshtml.cs
jrmccannon: There's not an enumeration problem. However, someone could post to the endpoint and spam create apps. I can put a check in there to make sure the app doesn't exist already.

jonashendrickx: I don't know, could avoid some paranoia for targeted attacks against specific `appid`

src/AdminConsole/Pages/Initialize.cshtml.cs
jonashendrickx: alright. you could even do a polling mechanism to auto-continue. It could also make the setup process for self-hosting more user-friendly

src/AdminConsole/Pages/Initialize.cshtml.cs
jrmccannon: I converted this over to blazor...for the practice of using blazor. We don't accept an appId so they can't really target an attack against that as far as I'm aware of.

src/AdminConsole/Pages/Initialize.cshtml
jrmccannon: I've removed this part.

src/AdminConsole/Pages/Initialize.cshtml
jrmccannon: It did some highlighting in rider...but not anywhere else.

src/Service/MagicLinks/MagicLinkService.cs
```diff
@@ -87,22 +87,26 @@ public async Task SendMagicLinkAsync(MagicLinkTokenRequest request)
         await mailProvider.SendAsync(new MailMessage
         {
             To = [request.EmailAddress.ToString()],
-            Subject = "Verify Email Address",
-            TextBody = $"Please click the link or copy into your browser of choice to log in: {link}. If you did not request this email, it is safe to ignore.",
+            Subject = $"Sign In to {link.Host}",
```
abergs, line 90: nit: Perhaps lowercase i?

jonashendrickx (approved):

jrmccannon (approved):
We did good work.

PR #422, opened 2024-02-16
Upgrade to .NET 8.0.2

jonashendrickx:
Requirement:
- Install the latest SDK: https://devblogs.microsoft.com/dotnet/february-2024-updates/
- Azure changes required: none

Side effects:
- IDE0005: Possible behavioral change regarding cleaning up `usings` when using `_Imports.razor`, causes other projects to be cleaned as well. Possibly expected functionality change: https://github.com/dotnet/format/issues/2119


abergs (changes requested):
After discussing this internally, we will wait a bit until going up to see if the `dotnet format` issues gets sorted out.

Blocking with a "Request changes" until then.

PR #423, opened 2024-02-16
PAS 370 - Add TTL

jrmccannon:
Adding a time to live property to SignInTokenRequest.  This will allow the caller to set a length of time that the token is valid for. It is measured in seconds and is defaulted to 120 seconds.


tests/Api.IntegrationTests/Endpoints/SignIn/SignInTests.cs
```diff
@@ -211,4 +212,42 @@ public async Task I_receive_a_sign_in_token_for_a_valid_user_id()
...
+        // Act
+        var verifySignInResponse = await client.PostAsJsonAsync("/signin/verify", new SignInVerifyDTO { Token = generateToken.Token, Origin = PasswordlessApi.OriginUrl, RPID = PasswordlessApi.RpId });
+
+        // Assert
+        verifySignInResponse.StatusCode.Should().Be(HttpStatusCode.Forbidden);
+        var problemDetails = await verifySignInResponse.Content.ReadFromJsonAsync<ProblemDetails>();
+        problemDetails.Should().NotBeNull();
+        problemDetails!.Extensions["errorCode"]!.ToString().Should().Be("expired_token");
+        problemDetails.Status.Should().Be(403);
+        problemDetails.Title.Should().Be("The token expired 00:00:10 ago.");
```
Tyrrrz, line 251: I know this is not part of this PR, but just thinking out loud (no need to change this in this PR or at all):

`00:00:10 ago` is not very natural to read. Maybe we should normalize the value to seconds/minutes? Could either use [Humanizer](https://github.com/Humanizr/Humanizer) or just write a simple extension method.

src/Service/Models/SigninTokenRequest.cs
```diff
@@ -1,6 +1,10 @@
 namespace Passwordless.Service.Models;
 
-public class SigninTokenRequest(string userId) : RequestBase
+public class SigninTokenRequest(string userId, int? timeToLive = null) : RequestBase
 {
     public string UserId { get; } = userId;
+    /// <summary>
+    /// Time to live is the number of seconds the token has before it expires.
+    /// </summary>
+    public int? TimeToLive { get; } = timeToLive;
```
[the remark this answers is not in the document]
jrmccannon: I disagree. This is the request model. In order for it to be a timespan, the user would have to post "00:00:13" as a string for a 13 second timespan. IMO, it would be much easier to say 13 seconds as a time to live. But I could be convinced otherwise.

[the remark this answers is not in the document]
Tyrrrz: We can make a converter that supports both `13` and `00:00:13` as input. The issue is that sooner or later `obj.TimeToLive` will surface somewhere in code, and it will not be clear what units it operates with.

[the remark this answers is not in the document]
jrmccannon: As an intermediate step, i'll create a DTO that takes the request object and converts it to a timespan. Later we can add that converter if we feel its needed.

tests/Api.IntegrationTests/Endpoints/SignIn/SignInTests.cs
jrmccannon: We already have it in the project so I added it.

src/Service/Models/SigninTokenRequest.cs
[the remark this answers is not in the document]
abergs: I agree that the API is nicer with seconds. I think the intermediate step of converting to span seems reasonable.

src/Service/Fido2Service.cs
```diff
@@ -292,8 +294,8 @@ public async Task<string> CreateSigninToken(SigninTokenRequest request)
         {
             Success = true,
             UserId = request.UserId,
-            Timestamp = DateTime.UtcNow,
-            ExpiresAt = DateTime.UtcNow.AddSeconds(120),
+            Timestamp = _timeProvider.GetUtcNow().DateTime,
+            ExpiresAt = _timeProvider.GetUtcNow().DateTime.Add(request.TimeToLiveTimeSpan),
```
abergs, line 298: What's the default value?

jrmccannon: default value is set in the SigninTokenRequest model. It is 120 seconds currently

src/Service/Models/MagicLinkTokenRequest.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public MagicLinkTokenRequest(string userId, MailAddress emailAddress, string linkTemplate)
+        : this(userId, emailAddress, linkTemplate, new TimeSpan(0, 15, 0)) { }
+    public MagicLinkTokenRequest(string userId, MailAddress emailAddress, string linkTemplate, int timeToLive)
+        : this(userId, emailAddress, linkTemplate, new TimeSpan(0, 0, timeToLive)) { }
+
+    public string UserId { get; } = userId;
+    public MailAddress EmailAddress { get; } = emailAddress;
+    public string LinkTemplate { get; } = linkTemplate;
+    public TimeSpan TimeToLive { get; } = timeToLive;
```
Tyrrrz, line 22: To simplify, consider making `TimeSpan timeToLive` nullable.

Then update the property like so:

```csharp
public TimeSpan TimeToLive { get; } = timeToLive ?? TimeSpan.FromMinutes(15);
```

This will make the default more obvious, will simplify the constructors, and will remove the need for ternary in `SendMagicLinkRequest.ToDto()`.

src/Service/Models/MagicLinkTokenRequest.cs
```diff
@@ -0,0 +1,24 @@
...
+/// <param name="userId">User identifier for the recipient of the magic link.</param>
+/// <param name="emailAddress">Email address for the intended user.</param>
+/// <param name="linkTemplate">Template used for creating the magic link. Token template string will be swapped with actual token.</param>
+/// <param name="timeToLive">Time span the magic link will be valid for. Default lifespan is 15 minutes.</param>
+public class MagicLinkTokenRequest(string userId, MailAddress emailAddress, string linkTemplate, TimeSpan timeToLive) : RequestBase
+{
+    public MagicLinkTokenRequest(string userId, MailAddress emailAddress, string linkTemplate)
+        : this(userId, emailAddress, linkTemplate, new TimeSpan(0, 15, 0)) { }
+    public MagicLinkTokenRequest(string userId, MailAddress emailAddress, string linkTemplate, int timeToLive)
+        : this(userId, emailAddress, linkTemplate, new TimeSpan(0, 0, timeToLive)) { }
```
Tyrrrz, line 17: Don't use `new TimeSpan(...)` like that. If `timeToLive` is above 60, then the above code will throw. Use `TimeSpan.FromSeconds(...)` etc.

Also, I'd argue we don't need the extra constructor for `int timeToLive` since we do the conversion in `SendMagicLinkRequest.ToDto()` already, and we can just do `TimeSpan.FromSeconds(...)` there, making things simpler.

src/Service/Models/SigninTokenRequest.cs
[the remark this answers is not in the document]
Tyrrrz: In that case, how about this:

```csharp
[JsonPropertyName("timeToLive")]
[Obsolete("This property is only used for serialization")]
public int? TimeToLiveSeconds { get; }

[JsonIgnore]
public TimeSpan TimeToLive { get; } = TimeSpan.FromSeconds(TimeToLiveSeconds ?? 120);
```

src/Service/Models/SigninTokenRequest.cs
```diff
@@ -1,6 +1,12 @@
...
+
+    public TimeSpan TimeToLiveTimeSpan => new(0, 0, 0, TimeToLive ?? 120);
```
Tyrrrz, line 11: This needs to call `FromSeconds`

Tyrrrz: Also, we have the default value of `120` specified now in two different places:

- here
- `MagicLinkTokenRequest`

src/Service/Validation/RegisterTokenValidator.cs
```diff
@@ -30,4 +31,18 @@ public static void ValidateAttestation(RegisterToken token, IFeaturesContext fea
             throw new ApiException("attestation_not_supported_on_plan", "Attestation type not supported on your plan", 400);
         }
     }
+
+    public static void Validate(this Token token, TimeProvider timeProvider)
```
Tyrrrz, line 35: If we don't want this method to be a part of `FidoService` and receive `TimeProvider` as a contextual parameter, then this static method can just simply take a `DateTimeOffset now` parameter instead.

src/Service/Models/SigninTokenRequest.cs
jrmccannon: The MagicLinkTokenRequest and SignInTokenRequest are two different requests that do have different. Should be okay to keep them separately defined. Did we want to make them the same length at this point?

Tyrrrz: Sorry, maybe I misunderstood then. I thought they were meant to be the same.

abergs: @jrmccannon Is correct in that they are separate and I think it makes sense to define them separately.

I do wonder if 15 minutes is too short for an email - not sure?
For MGAT I think at least 15 minutes seems reasonable for a default.

Tyrrrz (approved):

src/Service/Models/SigninTokenRequest.cs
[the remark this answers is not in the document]
jonashendrickx: But why is that timespan assuming a default value? Seems odd.

src/Service/Extensions/ConversionFunctions.cs
```diff
@@ -0,0 +1,9 @@
+namespace Passwordless.Service.Extensions;
+
+public static class ConversionFunctions
+{
+    public static TimeSpan? GetNullableTimeSpanFromSeconds(this int? seconds) =>
```
Tyrrrz, line 5: Nit: if you change the signature to `public static TimeSpan? ToTimeSpanFromSeconds(this int seconds)` (non nullable), then you can update the callsite like this:

```csharp
var timespan = timeToLive?.ToTimeSpanFromSeconds();
```

Which is subjectively a bit cleaner because it doesn't look like you're calling a method on a potentially null reference.

Tyrrrz (approved):

src/Service/Models/SigninTokenRequest.cs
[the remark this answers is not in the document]
jrmccannon: Should this be on a new type farther down?

Something like this:
```
using Passwordless.Service.Models;

namespace Passwordless.Service.MagicLinks.Models;

public class MagicLinkToken : VerifySignInToken
{
    private static readonly TimeSpan DefaultTimeToLive = TimeSpan.FromHours(1);

    private static TimeSpan GetTimeToLive(MagicLinkTokenRequest request) => request.TimeToLive ?? DefaultTimeToLive;
    
    public static MagicLinkToken GetInstance(MagicLinkTokenRequest request, DateTimeOffset now) => new()
    {
        Success = true,
        UserId = request.UserId,
        Timestamp = now.UtcDateTime,
        ExpiresAt = now.UtcDateTime.Add(GetTimeToLive(request)),
        TokenId = Guid.NewGuid(),
        Type = "magic_link",
        RpId = request.RPID,
        Origin = request.Origin
    };
}
```

That way the default value is encapsulated by the actual token it is for, instead of being a part of the request.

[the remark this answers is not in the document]
abergs: Seems like a bit too much indirection for my "keep it simple stupid" taste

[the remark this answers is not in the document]
abergs: I think Oleksii's approach is fine

src/Service/Extensions/ConversionFunctions.cs
```diff
@@ -0,0 +1,6 @@
+namespace Passwordless.Service.Extensions;
+
+public static class ConversionFunctions
+{
+    public static TimeSpan? ToTimeSpanFromSeconds(this int seconds) => TimeSpan.FromSeconds(seconds);
```
Tyrrrz, line 5: One last thing (sorry I must've missed this in my own code snippet)

```suggestion
    public static TimeSpan ToTimeSpanFromSeconds(this int seconds) => TimeSpan.FromSeconds(seconds);
```

Tyrrrz (approved):

src/Service/Validation/RegisterTokenValidator.cs
```diff
@@ -30,4 +31,16 @@ public static void ValidateAttestation(RegisterToken token, IFeaturesContext fea
             throw new ApiException("attestation_not_supported_on_plan", "Attestation type not supported on your plan", 400);
         }
     }
+
+    public static void Validate(this Token token, DateTimeOffset now)
```
jonashendrickx, line 35: Passing `now` as a parameter seems anti-pattern. Maybe find a way to make it work by injecting TimeProvider.

abergs: I should add the context that we do use Timeprovider, but since this is an extension method you can't inject, just use arguments. And it made less sense to send a timprovider as an argument compared to just a time (that itself is fetched from the timeprovider)

jonashendrickx: But if that gets too complicated it would mean the structure of the code is often wrong elsewhere. You can even simplify the use of `TimeProvider` by injecting an interface and just abstract it.

jrmccannon: Each way has its trade-offs. 

By providing TimeProvider, we're controlling the inputs to the method (since its an abstraction around Time) and allows the method to encapsulate what a valid token is.

By providing `DateTimeOffset` called `now`, we're essentially doing the same thing except using the plain `DateTimeOffset` (which could be anything). 

TimeProvider kind of builds the "system" into the method and `now` conveys the intent...just allows any `DateTimeOffset` to be passed in.

PR #424, opened 2024-02-16
PAS 371 - Change token template

jrmccannon:
Characters such as `<>`, `{}`, and `[]` are frequently escaped by frameworks. Switching from `<token>` to `$TOKEN` for the token template will hopefully help keep this feature painless.

### DIFF
--- a/src/Service/MagicLinks/MagicLinkService.cs
+++ b/src/Service/MagicLinks/MagicLinkService.cs
@@ -80,7 +80,7 @@ public class MagicLinkService(
         await EnforceQuotaAsync(dto);
 
         var token = await fido2Service.CreateSigninToken(new SigninTokenRequest(dto.UserId));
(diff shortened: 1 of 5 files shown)


Tyrrrz:
Can we do `$TOKEN` instead?

jonashendrickx (dismissed):

jrmccannon:
> Can we do `$TOKEN` instead?

@abergs did you see any issue with using `$TOKEN`? What library were you using that was getting rid of the `<>`?

abergs:
`$TOKEN` works well for me. This was the method and I confirmed it worked with $token too: https://github.com/bitwarden/passwordless-server/pull/421/files#diff-18b9f716b8cd4e9c2f632e3b88bc1870edc66c0537e99afd523c77b30eb7e23cR38

abergs (approved):

PR #439, opened 2024-02-25
[PAS289][Defect] Feedback no longer a confirmation step when deleting user credentials

jonashendrickx:
Using the confirmation dialog created by Anders to show the modal. To be able to use the reusable Razor component `ConfirmEditForm`, it was required to migrate the partial view to a Razor component as well.

The form submission is intercepted. When confirming, the form submission is allowed to execute. Upon successfully deleting the credential, the page is refreshed.

Added additional validation to ensure the credentialId is a 'base64url' encoded string.

**The alternative solution** would be to apply the attributes to the form and the button directly. But you would lose type safety. If the behavior happens to be altered, we'd have to hunt down every single usage of the confirmation modal.


src/AdminConsole/Pages/App/Credentials/User.cshtml.cs
```diff
@@ -22,24 +21,16 @@ public class UserModel : PageModel
...
+        Credentials = await _passwordlessClient.ListCredentialsAsync(UserId);
         Aliases = await _passwordlessClient.ListAliasesAsync(UserId);
     }
 
-    public async Task<IActionResult> OnPost(string token)
-    {
-        var res = await _passwordlessClient.VerifyTokenAsync(token);
-        return new JsonResult(res);
-    }
-
```
abergs, line 42: question: Why was this removed? Is it no longer needed?

jonashendrickx: It was not used at all. Might have been a left-over from when the playground was developed. I'll double check again.

src/AdminConsole/Components/Shared/Credentials.razor.cs
```diff
@@ -1,52 +1,83 @@
...
+using Microsoft.AspNetCore.Components;
+using Passwordless.AdminConsole.Helpers;
 using Passwordless.Common.Extensions;
+using Passwordless.Common.Validation;
 
-namespace Passwordless.AdminConsole.Pages.Shared;
+namespace Passwordless.AdminConsole.Components.Shared;
 
-public sealed class CredentialsModel
+public partial class Credentials : ComponentBase
```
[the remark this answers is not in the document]
jonashendrickx: If you split the `@code` section into a separate file from `Credentials.razor` it would be named `Credentials.razor.cs`. But you have to make the class partial: https://learn.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-8.0#partial-class-support

src/AdminConsole/Pages/App/Credentials/User.cshtml.cs
jonashendrickx: @abergs No other references are found. Seems ok to go

src/AdminConsole/Components/Shared/Credentials.razor.cs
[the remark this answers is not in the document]
abergs: OK

abergs (approved):

PR #440, opened 2024-02-26
Fix validation for application creation

jonashendrickx:


src/AdminConsole/Pages/Organization/CreateApplication.cshtml.cs
```diff
@@ -166,13 +166,13 @@ private async Task InitializeAsync()
...
-        [Required, MaxLength(60)]
+        [Required, MaxLength(60), MinLength(3), RegularExpression("^[a-zA-Z]{1}[a-zA-Z0-9 ]{2,59}$")]
         public string Name { get; set; }
 
-        [Required, MaxLength(62)]
+        [Required, MaxLength(62), MinLength(3), RegularExpression("^[a-z]{1}[a-z0-9]{2,61}$")]
         public string Id { get; set; }
 
-        [Required, MaxLength(120)]
+        [Required, MaxLength(120), MinLength(3), RegularExpression("^[a-zA-Z]{1}[a-zA-Z0-9 ]{2,119}$")]
```
abergs, line 175: Issue: I think we need to be more allowing in the description regexes than az09.

E.g. Description probably needs to allow commas, dashes, punctuation.

PR #443, opened 2024-02-27
Sanitize application name and description

jonashendrickx:
- We have no validation on the `appId` when an application is being created. For the validation to work on the route parameter, I needed to wrap it inside another object.
- Removes 'https://', 'http://', and HTML tags from application name and description.

### DIFF
--- a/src/AdminConsole/Pages/Organization/CreateApplication.cshtml.cs
+++ b/src/AdminConsole/Pages/Organization/CreateApplication.cshtml.cs
@@ -33,8 +33,7 @@ public class CreateApplicationModel : PageModel
         IApplicationService applicationService,
         IDataService dataService,
         IPasswordlessManagementClient managementClient,
-        IOptionsSnapshot<BillingOptions> billingOptions,
-        ISharedBillingService billingService)
+        IOptionsSnapshot<BillingOptions> billingOptions, ISharedBillingService billingService)
     {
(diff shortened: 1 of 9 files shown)


src/AdminConsole/Pages/Organization/CreateApplication.cshtml.cs
```diff
@@ -166,14 +167,38 @@ private async Task InitializeAsync()
...
+        [Required, MaxLength(60), MinLength(3)]
+        public string Name
+        {
+            get
+            {
+                return _name;
+            }
+            set
+            {
+                _name = HtmlSanitizer.Sanitize(value);
```
Tyrrrz, line 181: Shouldn't we send `Bad Request` instead of silently changing user input?

jonashendrickx: Because it is a page handler that either returns the same page or redirects to a new one. Returning a bad request would trigger the exception handler and redirect to the error page.

What you probably mean is you want to validate the form input, compare sanitized with unsanitized input and invalidate the form.

Tyrrrz: Yes, sorry. Forgot it was in AC. Essentially I'm wondering if communicating an error is better than silently fixing it.

PR #447, opened 2024-02-27
Dynamic matrix cases for the `deploy` job

Tyrrrz:
Currently, the job runs on all environments and just selectively skips the deployment step if the trigger event is not applicable (e.g. `push` for `prod` instead of `release`). This is fine, but it results in all instances of the `deploy` job marked as having completed, even if they didn't actually do anything.

As a workaround, I've changed the matrix to include an `is-release` parameter. It will be applied to all matrix cases like shown below.

- Workflow triggered by a release event:
  - Generated matrix:
    - `environment: devtest` / `is-release: true`
	- `environment: qa` / `is-release: true`
	- `environment: prod` / `is-release: true`
  - Exclusion on `is-release: false` does not apply since all cases are `true`
  - Final matrix is the same as the generated matrix
  - Deployment will happen to all environments 

- Workflow triggered by a non-release event (e.g. normal commit to `main`):
  - Generated matrix:
    - `environment: devtest` / `is-release: false`
	- `environment: qa` / `is-release: false`
	- `environment: prod` / `is-release: false`
  - Exclusion on `is-release: false` applies to the last two cases
  - Final matrix:
    - `environment: devtest` / `is-release: false`
  - Deployment only happens to `devtest`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -182,6 +182,8 @@ jobs:
 
   # Dispatches a separate deployment workflow in a private repository
   deploy:
+    if: ${{ github.event_name != 'pull_request' }}
+  
     needs:
       - version
       - format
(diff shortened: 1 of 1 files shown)


Tyrrrz:
Simplified the changes even further

jrmccannon (approved):

PR #449, opened 2024-02-28
No forbidden content validation

jonashendrickx:
### Ticket

n/a

### Description

Validates whether a property contains HTML tags, `http://` or `https://`.

### Shape

- Imported HtmlSanitizer in Common project.
- Abstracted HtmlSanitizer is used in validation attribute

### Screenshots

image

### Checklist
- [ ] __
- [ ] __

- [ ] __
- [ ] __


src/Common/Serialization/HtmlSanitizer.cs
```diff
@@ -0,0 +1,22 @@
+using System.Text;
+using Ganss.Xss;
+
+namespace Passwordless.Common.Serialization;
+
+public static class HtmlSanitizer
+{
+    private static readonly Ganss.Xss.HtmlSanitizer _htmlSanitizer = new(new HtmlSanitizerOptions
```
jrmccannon, line 8: this should be `HtmlSanitizer`

src/Common/Validation/NoForbiddenContentAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+        if (!(value is string))
+        {
+            throw new ArgumentException("This attribute can only be applied to strings.");
+        }
+
+        var valueString = (string)value;
+        var sanitizedValue = HtmlSanitizer.Sanitize(valueString);
+        return sanitizedValue == valueString;
```
jrmccannon, line 24:
```suggestion
        if (value is not string valueString)
        {
            throw new ArgumentException("This attribute can only be applied to strings.");
        }

        var sanitizedValue = HtmlSanitizer.Sanitize(valueString);
        return sanitizedValue == valueString;
```

This should be equivalent and a little easier to read.

jrmccannon (commented):
I'll let another person look at this to approve it, but here are my two suggestions.

src/Common/Serialization/HtmlSanitizer.cs
abergs: @jrmccannon Could you clarify your comment? I'm not following.

jrmccannon: Naming conventions from .editorconfig says it should be `HtmlSanitizer` instead of `_htmlSanitizer`

jonashendrickx: You cannot name it `HtmlSanitizer` because it collides with the class' name, I've just renamed it to `Instance`.

abergs: OK to me

abergs (approved):
Looks good

PR #450, opened 2024-02-28
Add some tests

pooya1380m:
Added Tests to SharedManagementServiceTests for CreateApiKeyAsync method

and I like to know what should I use instead of 

https://github.com/bitwarden/passwordless-server/blob/1f47bf111e4f4e6bb2aaee632006219e8cd135b8/src/Service/Storage/Ef/DbTenantFactory.cs#L14


jonashendrickx:
What do you mean you want to know what you'd like instead of the factory method?

pooya1380m:
> What do you mean you want to know what you'd like instead of the factory method?

 [Obsolete("Use injected `ITenantStorage` instead when not calling from `ManagementKey` scope.")]
 its obsolete

jonashendrickx:
> > What do you mean you want to know what you'd like instead of the factory method?
> 
> [Obsolete("Use injected `ITenantStorage` instead when not calling from `ManagementKey` scope.")] its obsolete

See my comment above :-)

pooya1380m:
> > > What do you mean you want to know what you'd like instead of the factory method?
> > 
> > 
> > [Obsolete("Use injected `ITenantStorage` instead when not calling from `ManagementKey` scope.")] its obsolete
> 
> See my comment above :-)

thanks

pooya1380m:
@jonashendrickx 
can you review this?

jonashendrickx:
> @jonashendrickx can you review this?

Can you run `dotnet format` in your working directory and push again?

tests/Service.Tests/Implementations/SharedManagementServiceTests.cs
```diff
@@ -414,5 +429,97 @@ public async Task ListApiKeysAsync_Returns_ExpectedResult()
...
+        // act
+        var actual = await _sut.CreateApiKeyAsync(appId, createPublicKey);
+
+        // assert
+        actual.ApiKey[..13]
+            .Should()
+            .Be("appId:public:");
+
+        storageMock.Verify(x => x.StoreApiKeyAsync(It.IsAny<string>(), It.IsAny<string>(),
+            It.Is<string[]>(p => p == scopes.Select(s => s.GetValue()).ToArray())), Times.Once);
```
pooya1380m, line 457: @jonashendrickx  it don't work I don't know why

pooya1380m:
> > @jonashendrickx can you review this?
> 
> Can you run `dotnet format` in your working directory and push again?
> 
> Looks good overall, you could theoretically also verify you're calling `StoreApiKey` with the parameters you expect to be passed. I've gotten bitten often enough by mappings that were wrong.
> 
> It could look like:
> 
> `storageMock.Verify(x => x.StoreApiKeyAsync(It.Is<string>(p => p == pk), It.Is<string>(p => p == publicKey), It.Is<PublicKeyScopes[]>(p => p == scopes)), Times.Once);`

@jonashendrickx 

I don't have the pk and publicKey in arrange level of test, the guid is generated inside of SetupApiKeyAsync method 
how can I have them as expected ?

I wrote it like this 

```c# 
 storageMock.Verify(x => x.StoreApiKeyAsync(It.IsAny<string>(), It.IsAny<string>(),
            It.Is<string[]>(p => p == scopes.Select(s => s.GetValue()).ToArray())), Times.Once);
```
but it give this error " Expected invocation on the mock once, but was 0 times" I don't know why
can you check this

PR #451, opened 2024-02-28
Switch from `DateTimeOffest` to `DateTime` in EF models

Tyrrrz:
### Ticket

N/A

### Description

As it turns out, EF Core provider for SQLite does not support `DateTimeOffset` comparisons, so queries like the one below fail:

```csharp
var from = timeProvider.GetUtcNow() - window;
return await db.DispatchedEmails
    // CreatedAt is of type DateTimeOffset
    .Where(x => x.CreatedAt >= from)
    .ToArrayAsync();
```

Interestingly, even though SQLite does not have any native support for dates (no `datetime` or similar column type), EF Core is able to handle the above query just fine when using regular `DateTime` on the CLR type. Why it doesn't work with `DateTimeOffest` is a mystery.

While you can provide a custom `ValueConverter` and `ValueComparer` inside `OnModelCreating`, this doesn't ultimately help with `>`/`<` comparisons that's used in the query above.

https://docs.microsoft.com/en-us/ef/core/providers/sqlite/limitations#query-limitations

### Shape

This changes the affected property from `DateTimeOffset` to `DateTime`. Migrations included will change the backing column type the MSSQL from `datetimeoffset` to `datetime2`.

### Screenshots

N/A

### Checklist

- [X] I ran the existing tests against both MSSQL and SQLite and both worked

N/A

### DIFF
--- a/src/Service/Migrations/Mssql/20240228181428_DateTimeOffsetToDateTime.Designer.cs
+++ b/src/Service/Migrations/Mssql/20240228181428_DateTimeOffsetToDateTime.Designer.cs
@@ -3,6 +3,7 @@ using System;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Migrations;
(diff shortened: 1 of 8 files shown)


jrmccannon (approved):
Sadly this is something we had to do. @abergs for visibility.

PR #463, opened 2024-03-05
Temporarily disable rate limiting in magic links endpoint

Tyrrrz:
### Ticket

N/A

### Description

This temporarily removes rate limits from the magic links endpoint so that it's a little easier to test the monthly quotas on the production environment. This is not a longterm-viable solution but we need something that would unblock us quickly.

### Shape

N/A

### Screenshots

N/A

### Checklist

N/A


abergs (approved):
Seems format failed @Tyrrrz

PR #465, opened 2024-03-06
PAS-148 | Data export

jonashendrickx:
### Ticket

- Closes [PAS-148](https://bitwarden.atlassian.net/browse/PAS-148)

### Description

We need to be able to export individual applications with all their relevant data, so customers can migrate to self-hosted instances.

### Shape

Given applications with either a lot of users or enterprise applications with event logging enabled can end up having a lot of records.

Considerations:
- We need to be able to enable paging easily at a later stage or possibly offload the data export process using a serverless approach.
- CSV:
  - RFC-4180 compliant
  - Easy to enable paging later
  - CsvHelper has a lot of capable functionality included to help us achieve paging easily later.
  - UTF-8 encoded

Remarks:
- Wasn't meant to include paging at this stage.
- Wasn't meant to include encryption at this stage.

Split to one file per entity type. Every file is a csv document compliant with RFC-4180. Splitting to a file per record makes it also easier theoretically to enable paging at a later stage when a file grows to a certain size.

`POST /backup/schedule`: Schedules a new backup job.

`GET /backup/jobs`: Retrieves a list of backup jobs with their relevant status past or present.

Example response:
```
[
{  "jobId": "guid", "status": "pending", "createdAt": "ISO-8160", "lastUpdatedAt": "ISO-8160" },
{  "jobId": "guid", "status": "failed", "createdAt": "ISO-8160", "lastUpdatedAt": "ISO-8160" },
{  "jobId": "guid", "status": "running", "createdAt": "ISO-8160", "lastUpdatedAt": "ISO-8160" },
{  "jobId": "guid", "status": "completed", "createdAt": "ISO-8160", "lastUpdatedAt": "ISO-8160" },
]
```

The data stored is stored as a UTF-8 encoded byte array.

```
public class ArchiveJob : PerTenant
{
    public Guid Id { get; set; }

    public AccountMetaInformation Application { get; set; }

    public DateTime CreatedAt { get; set; }

    public DateTime? UpdatedAt { get; set; }

    public JobStatus Status { get; set; } = JobStatus.Pending;

    public List<Archive> Archives { get; set; } = new();
}
```

```
public class Archive : PerTenant
{
    public Guid Id { get; set; }

    /// <summary>
    /// The identifier of the backup job that created this archive.
    /// </summary>
    public Guid JobId { get; set; }

    public DateTime CreatedAt { get; set; }

    public Type? Entity { get; set; }

    [MaxLength(100 * 1024 * 1024, ErrorMessage = "Data cannot be larger than 100MB.")]
    public byte[] Data { get; set; }

    public AccountMetaInformation Application { get; set; } = null!;
}
```

The upload process would likely involve a user to create a new organization, and import an old app by uploading the exported files. The restoration would only be able to start if the schema's match and when all documents were successfully uploaded.

**ApplicationEvents**

An additional migration for `ApplicationEvents` had to be included, as it was not inheriting from `PerTenant` according to the conventions we had laid out for `DbTenantContext`. This was then essentially messing with the generics I had implemented in `BackupWorkerService`. This would have otherwise significantly impacted maintainability as well.

To prevent any unnecessary migrations from being executed which could result into data loss, the database column was mapped manually to its original value. Just the snapshot was essentially updated to reflect the mapping from the column name to the property (CLR).

### Screenshots

n/a

### Checklist
- [ ] Unit tests
- [ ] Integration tests

- [ ] Secured endpoints with secret key.
- [ ] Sanitization for macro's was ignored, given these backups are not meant to be opened in a program like Microsoft Excel.
- [ ] In this implementation, a DoS attack would be possible with very large customers.

[PAS-148]: https://bitwarden.atlassian.net/browse/PAS-148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Service/Backup/BackupWorkerService.cs
```diff
@@ -0,0 +1,94 @@
...
+        await BackupEntityAsync<Authenticator>(job.Id, job.Tenant);
+        await BackupEntityAsync<AliasPointer>(job.Id, job.Tenant);
+        await BackupEntityAsync<EFStoredCredential>(job.Id, job.Tenant);
+        await BackupEntityAsync<ApplicationEvent>(job.Id, job.Tenant);
+        await BackupEntityAsync<PeriodicCredentialReport>(job.Id, job.Tenant);
+        await BackupEntityAsync<PeriodicActiveUserReport>(job.Id, job.Tenant);
+    }
+
+    private async Task BackupEntityAsync<TEntity>(Guid groupId, string tenant) where TEntity : PerTenant
+    {
```
jonashendrickx, line 77: @abergs & @jrmccannon For this I required to make `ApplicationEvent` inherit from `PerTenant`.

src/Service/Storage/Ef/DbGlobalContext.cs
```diff
@@ -87,9 +90,10 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
         modelBuilder.Entity<ApplicationEvent>(builder =>
         {
             builder.HasKey(x => x.Id);
```
jonashendrickx, line 92: This adds backwards compatibility, so we don't need to migrate any data, it just maps it to the new CLR property.

abergs:
Re-iterating earlier comment: Let's review this PR but do not merge it until we've shipped some of the things currently in `main` to prod.

src/Service/Backup/BackupBackgroundService.cs
```diff
@@ -0,0 +1,65 @@
...
+
+    public BackupBackgroundService(
+        IServiceProvider serviceProvider,
+        ILogger<BackupBackgroundService> logger)
+    {
+        _serviceProvider = serviceProvider;
+        _logger = logger;
+    }
+
+    protected async override Task ExecuteAsync(CancellationToken stoppingToken)
```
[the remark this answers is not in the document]
jonashendrickx: @jrmccannon That one was meant to run background jobs on fixed times during the day. This would just be a queue that takes the oldest one first to be processed.

PR #469, opened 2024-03-07
PAS-384 | Fix delete passkey bug once and for all

jonashendrickx:
### Ticket

- Closes [PAS-384](https://bitwarden.atlassian.net/browse/PAS-384)

### Description

We were unable to delete passkeys on the profile page, because it was trying to use the `IScopedPasswordlessClient`. We're now injecting the `IPasswordlessClient` or its derivative `IScopedPasswordlessClient` using the `PasswordlessClient` parameter on the`Credentials.razor` component. A second parameter was then created to pass the `UserId`.

Similarly the user is passed from the claim, rather than using the expensive `UserManager<ConsoleAdmin>` calling the database with an additional trip. The credentials for a user in a given app for a customer are passed using a query parameter `userId` this behavior was not changed.

### Shape

See above.

### Screenshots

n/a

### Checklist
- [ ] Test all 3 screens where credentials were present

- [ ] Reduced security impact where user id can be obtained from the claim directly for a logged-in admin

[PAS-384]: https://bitwarden.atlassian.net/browse/PAS-384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/AdminConsole/Components/Shared/Credentials.razor.cs
+++ b/src/AdminConsole/Components/Shared/Credentials.razor.cs
@@ -10,10 +10,6 @@ public partial class Credentials : ComponentBase
 {
     public const string RemoveCredentialFormName = "remove-credential-form";
 
-    /// <summary>
-    /// The list of credentials.
-    /// </summary>
-    [Parameter]
     public required IReadOnlyCollection<Credential> Items { get; set; }
 
     public IReadOnlyCollection<CredentialModel> GetItems()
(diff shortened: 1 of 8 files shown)


src/AdminConsole/Pages/Account/Profile.cshtml.cs
```diff
@@ -1,39 +1,9 @@
...
-
-    public async Task OnGet()
-    {
-        var userId = _userManager.GetUserId(HttpContext.User);
-        if (userId != null)
-        {
-            Credentials = await _client.ListCredentialsAsync(userId);
-        }
-    }
+    public string UserId => HttpContext.User.Claims.Single(c => c.Type == ClaimTypes.NameIdentifier).Value;
```
abergs, line 8: suggest, separate issue: It would be nice to have an extension method for this.

jonashendrickx: @abergs we apparently had one already

src/AdminConsole/Components/Shared/Credentials.razor.cs
```diff
@@ -46,11 +42,18 @@ public IReadOnlyCollection<CredentialModel> GetItems()
     [Parameter]
     public bool HideDetails { get; set; }
 
+    [Parameter]
+    public required IPasswordlessClient PasswordlessClient { get; set; }
```
abergs, line 46: question: Why doesn't it work with the injected one, why do we need to pass this as a parameter?

abergs (approved):
shape, tech: LGTM

src/AdminConsole/Components/Shared/Credentials.razor.cs
jonashendrickx: This component is used on 3 pages in total:

- /account/useronboarding: uses `IPasswordlessClient`
- /account/profile: uses `IPasswordlessClient`
- /app/<appid>/credentials/list: uses `IScopedPasswordlessClient` (which replaces the ApiSecret from admin app with the one from the customer's app)

PR #474, opened 2024-03-12
Implement value comparers for the array conversions

jonashendrickx:
### Ticket

n/a

### Description

We are currently converting text fields into arrays with Entity Framework, but we are not implementing value comparers. This could impact change tracking: https://learn.microsoft.com/en-us/ef/core/modeling/value-comparers?tabs=ef5

### Shape

n/a

### Checklist
- [ ] Unit testing

### DIFF
--- a/src/Service/Storage/Ef/DbGlobalContext.cs
+++ b/src/Service/Storage/Ef/DbGlobalContext.cs
@@ -1,11 +1,13 @@
 using System.Text.Json;
 using Fido2NetLib.Objects;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.ChangeTracking;
 using Passwordless.Common.Constants;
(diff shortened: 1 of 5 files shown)


src/Service/Storage/Ef/DbGlobalContext.cs
```diff
@@ -36,7 +38,8 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             b.HasKey(x => new { x.Tenant, x.DescriptorId });
             b.Property(x => x.DescriptorTransports).HasConversion(
```
jrmccannon, line 39: We don't have to change anything here. But next time we are storing multiple compile time constants (such as an enum) to the database, we should consider using the `[Flags]` attribute which will store the value as an integer.  Each enum would then have a binary representation for values.

jrmccannon (approved):

PR #476, opened 2024-03-12
Update pull_request_template.md

abergs:
### Description

Adresses some feedback on the PR template.
Moves us to use bullet list instead of checkboxes and tries to simplify the template in terms of visual focus when creating a new PR (moving some comments to be  inline)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,21 +1,17 @@
 <!--
 Thank you for contributing! Please follow the steps below to help us process your PR quickly.
-- 📝 Use a meaningful title for the pull request.
+- 📝 Use a meaningful title for the pull request, e.g: "PAS-XXX | short pr description"
 - 💭 Write a clear description and share screenshots (if applicable) to help describe your change.
 - 🔍 Not all sections below will apply to you and are mostly for our internal team. It's okay to delete them if they are not applicable. 
 -->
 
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -37,9 +32,9 @@ Thank you for contributing! Please follow the steps below to help us process you
 
 ### Checklist
 I did the following to ensure that my changes were tested thoroughly:
-- [ ] __
-- [ ] __
+* __
+* __
```
Tyrrrz, line 36: I find that I ever only need one bullet point, so maybe it makes sense to make it the default

```suggestion
- __
```

.github/pull_request_template.md
```diff
@@ -37,9 +32,9 @@ Thank you for contributing! Please follow the steps below to help us process you
...
 
-I did the following to ensure that my changes did not introduce new security vulnerabilities:
-- [ ] __
-- [ ] __
+I did the following to ensure that my changes do not introduce security vulnerabilities:
+* __
+* __
```
Tyrrrz, line 40: I find that I ever only need one bullet point, so maybe it makes sense to make it the default

```suggestion
- __
```

jonashendrickx (approved):

jrmccannon (approved):

Tyrrrz (approved):

PR #478, opened 2024-03-13
Delete expired organization events at midnight UTC.

jonashendrickx:
### Description

Currently, the background services run at random times during the day depending on when the server has last restarted. This pull request makes that `organization delete background worker` runs at a fixed time during the day, 'midnight UTC'.

It will also return the amount of deleted records, for use with logging.

### Checklist
- [ ] Run once

### DIFF
--- a/src/AdminConsole/EventLog/EventDeletionBackgroundWorker.cs
+++ b/src/AdminConsole/EventLog/EventDeletionBackgroundWorker.cs
@@ -1,42 +1,35 @@
 using Passwordless.AdminConsole.EventLog.Loggers;
+using Passwordless.Common.Background;
 
 namespace Passwordless.AdminConsole.EventLog;
 
-public class EventDeletionBackgroundWorker : BackgroundService
+public class EventDeletionBackgroundWorker : BasePeriodicBackgroundService
 {
-    private readonly ILogger<EventDeletionBackgroundWorker> _logger;
     private readonly IServiceProvider _serviceProvider;
 
     public EventDeletionBackgroundWorker(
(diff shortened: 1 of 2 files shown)


src/AdminConsole/EventLog/EventDeletionBackgroundWorker.cs
```diff
@@ -1,42 +1,35 @@
...
-    private readonly ILogger<EventDeletionBackgroundWorker> _logger;
     private readonly IServiceProvider _serviceProvider;
 
     public EventDeletionBackgroundWorker(
-        ILogger<EventDeletionBackgroundWorker> logger,
-        IServiceProvider serviceProvider)
+        IServiceProvider serviceProvider,
+        TimeProvider timeProvider,
+        ILogger<EventDeletionBackgroundWorker> logger)
+        : base(new TimeOnly(0), TimeSpan.FromDays(1), timeProvider, logger)
```
jrmccannon, line 14: Nit: Might be nice to have some `static readonly` "constants" that were a little more reader friendly. 

Thinking maybe a static class somewhere that had 
```
public static class TimeConstants 
{
    public static readonly TimeOnly Midnight = new TimeOnly(0);
}
```

Then when we see it, we instantly know "oh that's midnight."

jrmccannon (approved):
Small comment. Not a PR stopper.

src/AdminConsole/EventLog/EventDeletionBackgroundWorker.cs
jonashendrickx: I need to move more to make it worth it. There is a couple more to move.

PR #479, opened 2024-03-13
Refactor validation for registering tokens, setting alias and magic links

jonashendrickx:
### Description

Validation is currently a spaghetti nightmare that's all over the place. This moves certain validation to earlier to the controllers.

### Shape

n/a

### Checklist
- [ ] Checked integration tests
- [ ] Manual testing

### DIFF
--- a/src/Service/Models/AliasPayload.cs
+++ b/src/Service/Models/AliasPayload.cs
@@ -1,10 +1,15 @@
-﻿namespace Passwordless.Service.Models;
+﻿using System.ComponentModel.DataAnnotations;
+using Passwordless.Common.Validation;
 
-using Aliases = HashSet<string>;
+namespace Passwordless.Service.Models;
 
-public class AliasPayload
-{
-    public string UserId { get; set; }
-    public Aliases Aliases { get; set; }
-    public bool Hashing { get; set; } = true;
-}
+public record AliasPayload
+(
+    [Required(AllowEmptyStrings = false)]
+    string UserId,
+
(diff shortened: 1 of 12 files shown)


src/Service/Models/AliasPayload.cs
```diff
@@ -1,10 +1,15 @@
...
-{
-    public string UserId { get; set; }
-    public Aliases Aliases { get; set; }
-    public bool Hashing { get; set; } = true;
-}
\ No newline at end of file
+public record AliasPayload
+(
+    [Required(AllowEmptyStrings = false)]
+    string UserId,
```
jrmccannon, line 9: We should consider creating a `UserIdValidation` attribute that encompasses what a valid user id is for the API.  That way its clear that when you have an UserId in the API, it should have a specific shape and that shape will be defined in a single spot.

src/Service/Models/AliasPayload.cs
```diff
@@ -1,10 +1,15 @@
...
+
+    [MaxLength(10), MaxLengthCollection(250), RequiredCollection(AllowEmptyStrings = false)]
+    HashSet<string> Aliases,
```
jrmccannon, line 12: We should consider creating an `AliasCollectionValidation` or `AliasValidation` attribute that encompasses what a valid alias/collection of aliases is for the API. That way its clear that when you have an Alias in the API, it should have a specific shape and that shape will be defined in a single spot.

jrmccannon (commented):
This doesn't need to stop this PR, but left a few comments to consider.

jrmccannon (approved):

PR #481, opened 2024-03-14
PAS-202 | OpenAPI

jonashendrickx:
### Ticket

- Closes [PAS-202](https://bitwarden.atlassian.net/browse/PAS-202)

### Description

Generate the OpenAPI documentation to allow customers to write their own libraries.

In the TODO block below, I've left out some things, to avoid merge conflicts:

**TODO:**

- Later: We should add more descriptions to each endpoint individually.
- Later: Any additional custom validation attributes we create, need to be added to populate the schema properly.

`/swagger/index.html`

### Shape

- Registering the policies individually was quite cumbersome, and didn't allow me to identify the authorization scheme that was being used. So I couldn't set the parameter.
- Using Authorization requirements gives us more flexibility. I'm still verifying that the same claims are present in similar fashion to what we had before. https://learn.microsoft.com/en-us/aspnet/core/security/authorization/iard
- `ManagementKey` authentication scheme is ignored automatically from the OpenAPI documentation, while `ApiKey` and `ApiSecret` schemes are added automatically. They can still be ignored with specific claims, or ignore parameters.
- Disabling the `POST /signin/generate-token` endpoint was returning a 403 manually. This caused some issues. I've changed it to throw a proper `ApiException` so we're returned a `ProblemDetails` object accordingly. Given I removed the fallback authentication scheme which wasn't configured properly, it was stumbling over it. 
- There is support for nullable reference types and validation attributes.
- Performance improvements to check what policy was applied to a specific endpoint. We no longer have to populate all the policies during startup: [https://github.com/bitwarden/passwordless-server/pull/481/files#diff-29b41c5928b12a24942353d7b14013f675989d1b393ac2dc66b02a161346e365](https://github.com/bitwarden/passwordless-server/pull/481/files#diff-29b41c5928b12a24942353d7b14013f675989d1b393ac2dc66b02a161346e365)

### Screenshots

Authenticators have been completed a bit more with XML comments.

image

### Checklist
- [ ] Verify integration tests
- [ ] Inspect the generated documentation for anomalies

- [ ] Verify integration tests

[PAS-202]: https://bitwarden.atlassian.net/browse/PAS-202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Api/Authorization/EndpointConventionBuilderExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// </summary>
+    /// <param name="builder"></param>
+    /// <param name="scope">When provided, requires the `ApiKey` to have the specified scope.</param>
+    /// <typeparam name="TBuilder"></typeparam>
+    /// <returns></returns>
+    public static TBuilder RequirePublicKey<TBuilder>(this TBuilder builder, PublicKeyScopes? scope = null) where TBuilder : IEndpointConventionBuilder
+    {
+        var requirements = new List<IAuthorizationRequirement>
+        {
+            new ClaimsAuthorizationRequirement(CustomClaimTypes.AccountName, null),
```
jonashendrickx, line 39: We should have an `AccountName` claim for public and secret keys.

src/Api/Authorization/EndpointConventionBuilderExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+            new ClaimsAuthorizationRequirement(CustomClaimTypes.KeyType, new []{ Constants.PublicKeyType }),
+        };
+
+        if (scope.HasValue)
```
jonashendrickx, line 43: If a scope was required by an endpoint, add the `ClaimAuthorizationRequirement` to check the claim is present for the authenticated `ApiKey` or `SecretKey`.

src/Api/Program.cs
```diff
@@ -24,6 +26,23 @@
 
 WebApplicationBuilder builder = WebApplication.CreateBuilder(args);
 
+builder.Services.AddEndpointsApiExplorer();
+builder.Services.AddSwaggerGen(swagger =>
+{
+    swagger.DocInclusionPredicate((docName, apiDesc) =>
+    {
+        var policy = (AuthorizationPolicy?)apiDesc.ActionDescriptor.EndpointMetadata.SingleOrDefault(x => x.GetType() == typeof(AuthorizationPolicy));
```
jonashendrickx, line 34: Skip any endpoints protected by the `ManagementKey` authentication scheme.

src/Api/Program.cs
```diff
@@ -72,10 +91,10 @@
 
 services.AddProblemDetails();
 services
-    .AddAuthentication(Constants.Scheme)
```
jonashendrickx, line 75: This acts as a fallback, but was never working properly.

src/Api/Program.cs
```diff
@@ -72,10 +91,10 @@
...
+    .AddAuthentication()
     .AddCustomSchemes();
 
-services.AddAuthorization(options => options.AddPasswordlessPolicies());
```
jonashendrickx, line 78: We no longer need to map the policies.

src/Api/Api.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <UserSecretsId>passwordless-api</UserSecretsId>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
jonashendrickx, line 5: This generates a `passwordless.api.xml` file containing the XML comments.

Tyrrrz (approved):

Tyrrrz (approved):

PR #489, opened 2024-03-15
Add missing AppFeature records

jrmccannon:
### Ticket
No ticket.

### Description
With the replacement of magic links in Admin Console with the API, it is relying on the AppFeature record to exist and for the monthly quota to be set.  With it not set, AdminConsole can't send magic links.

### Shape
- Added the max quota to the intiialize steps for AdminConsole.
- Also added a migration that will add an AppFeature record with some default values to every app that currently doesn't have an AppFeature record.  This would be every application created before AppFeatures went live (June-August time frame).

### Screenshots
N/A

### Checklist
- Ran the script against my local database to make sure it executed and only updated the records it should have

- N/A


jonashendrickx (approved):

abergs:
Looks mostly good, see comment

PR #504, opened 2024-03-19
PAS-375 | Magic links <24h condition breaking self-hosting

jonashendrickx:
### Description

When you create a new self-hosting environment, you go through the '/initialize' page where a new organization and 'adminconsole' application will be created. However, when you then successfully proceed to the signup page, you essentially cannot register the actual organization that will hold all your applications. So basically the first 24 hours every customer will be blocked from doing anything at all with their self-hosting environment.

This condition will likely also cause a lot of problems during development if potential customers want to have multiple developers or admins work together during the first day.

### Shape

It would be better to adjust the enforcement quota based on application age. And gradually increase as time passes.

### Screenshots

n/a

### Checklist
- __

- __

### DIFF
--- a/src/AdminConsole/Components/Pages/Initialize.razor
+++ b/src/AdminConsole/Components/Pages/Initialize.razor
@@ -10,6 +10,7 @@
 @using Microsoft.AspNetCore.Identity
 @using Passwordless.AdminConsole.Models
 @using System.ComponentModel.DataAnnotations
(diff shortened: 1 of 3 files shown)


Tyrrrz:
> This condition will likely also cause a lot of problems during development if potential customers want to have multiple developers or admins work together during the first day.

Technically, the corresponding field in the database is an array, so we could fit more than one email in there.

PR #507, opened 2024-03-20
Allow AC to send magic links

jrmccannon:
### Ticket
- Closes [PAS-375](https://bitwarden.atlassian.net/browse/PAS-375)

### Description
AdminConsole is being restricted by how recently it was created in new self hosted instances. This checks if the requesting application is admin console by tenant name.

### Shape
Adds an `IsAdminConsole` check when enforcing the 24h limit quota.

### Checklist
- ran unit tests.

- ensured quota enforcement was kept in place.

[PAS-375]: https://bitwarden.atlassian.net/browse/PAS-375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Service/MagicLinks/MagicLinkService.cs
+++ b/src/Service/MagicLinks/MagicLinkService.cs
@@ -1,4 +1,3 @@
-using Microsoft.Extensions.Caching.Memory;
 using Passwordless.Common.MagicLinks.Models;
 using Passwordless.Common.Services.Mail;
 using Passwordless.Service.EventLog.Loggers;
@@ -10,12 +9,13 @@ namespace Passwordless.Service.MagicLinks;
 
 public class MagicLinkService(
     TimeProvider timeProvider,
-    IMemoryCache cache,
     ITenantStorage tenantStorage,
     IFido2Service fido2Service,
     IMailProvider mailProvider,
     IEventLogger eventLogger)
 {
+    private static readonly string[] AdminConsole = ["admin", "adminconsole"];
+    
     private async Task EnforceQuotaAsync(MagicLinkTokenRequest request)
     {
         var now = timeProvider.GetUtcNow();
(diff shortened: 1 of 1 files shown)


src/Service/MagicLinks/MagicLinkService.cs
```diff
@@ -64,6 +65,8 @@ private async Task EnforceQuotaAsync(MagicLinkTokenRequest request)
         }
     }
 
+    private static bool IsAdminConsole(AccountMetaInformation account) => AdminConsole.Contains(account.Tenant);
```
Tyrrrz, line 68: Suggestion: simplify to avoid additional noise with the field and keep all related logic collocated:

```suggestion
    private static bool IsAdminConsole(AccountMetaInformation account) =>
        string.Equals(account.Tenant, "admin", StringComparison.OrdinalIgnoreCase) ||
        string.Equals(account.Tenant, "adminconsole", StringComparison.OrdinalIgnoreCase);
```

Tyrrrz (approved):

src/Service/MagicLinks/MagicLinkService.cs
```diff
@@ -64,6 +63,10 @@ private async Task EnforceQuotaAsync(MagicLinkTokenRequest request)
         }
     }
 
+    private static bool IsAdminConsole(PerTenant account) =>
+        string.Equals(account.Tenant, "admin", StringComparison.OrdinalIgnoreCase)
+        || string.Equals(account.Tenant, "adminconsole", StringComparison.OrdinalIgnoreCase);
```
Tyrrrz, line 68: We should probably do something to prevent regular users from creating apps with these names? For example if `adminconsole` already exists and the user creates `admin`.

Tyrrrz (approved):

jonashendrickx (commented):
Still wouldn't work for local development. This is a very dirty workaround though. The API shouldn't have any concept of what the admin console is

PR #509, opened 2024-03-20
PAS-392 | Add rate limit bypass

Tyrrrz:
### Ticket

- Closes https://bitwarden.atlassian.net/browse/PAS-392

### Description

Adds a configuration flag `Flags:IsRateLimitBypassEnabled` that can be set to `true` to allow clients to bypass rate limits. If the request includes header `X-RateLimit-Bypass` set to `true`, the rate limiting policies are not applied to it. This is useful when testing in production-like environments, i.e. `devtest` and `qa`.

To finalize this feature, after merging the PR, we need to add the above configuration flag to their respective environments. @abergs

### Shape

See above.

### Checklist
- Regression tests passed

- I gave it a thought and it seems okay


withinfocus (commented):
Are these environments publicly-accessible? We're risking them if we allow this to not be controllable per caller. https://github.com/bitwarden/server/blob/main/perf/load/helpers/auth.js#L27 shows what we do for Bitwarden, and specific client IDs are used via environment variable injection to control this per-environment.

Tyrrrz:
> We're risking them if we allow this to not be controllable per caller. [bitwarden/server@`main`/perf/load/helpers/auth.js#L27](https://github.com/bitwarden/server/blob/main/perf/load/helpers/auth.js?rgh-link-date=2024-03-20T20%3A13%3A28Z#L27) shows what we do for Bitwarden, and specific client IDs are used via environment variable injection to control this per-environment.

Yeah that might be easier.

Do you have an example of how the client IDs are checked on the backend?

withinfocus:
> Do you have an example of how the client IDs are checked on the backend?

CloudOps configures them by setting `.env` entries for `IpRateLimitOptions__ClientWhitelist`, and then via https://github.com/bitwarden/server/blob/main/src/Identity/appsettings.json#L77 our rate limiter ignores those clients.

src/Api/appsettings.Development.json
```diff
@@ -1,4 +1,7 @@
 {
+  "Flags": {
+    "IsRateLimitBypassEnabled": true
```
abergs, line 3:
```suggestion
    "IsRateLimitBypassEnabled": ["qa-app-1","qa-app-2"]
```

I think this kind of pattern would solve what Matt described and also solves QA's requirements.

withinfocus:
I noticed the ASP.NET Core rate limiter is used here too, and if this service scales beyond one instance that system will essentially fail. https://github.com/cristipufu/aspnetcore-redis-rate-limiting can be used to get Redis backplane support like we use with Bitwarden, albeit using an older library before ASP.NET provided rate limiting (7+); we'd like to migrate.

Tyrrrz:
So is the suggested approach to avoid abuse the following?

- Set up some kind of client identification header
- Pass it through to the backend via `PasswordlessClient` or manual requests
- On the backend, configure secret client IDs that are "trusted"
- Compare if the IDs match and allow/disallow elevated permissions

If that's the case, maybe it's easier to just have some sort of secret "god mode" header instead?

- Set up the god mode passphrase as a secret on the backend (only on QA and devtest)
- Send it along as a separate header in addition to the `X-Bypass-RateLimit` header
- Compare the values and allow/disallow

withinfocus:
The "allow/disallow elevated permissions" is rather open-ended, but in this scope it's about applying rate-limiting or not. I like the approach of a client identifier because it can be used for many things depending on what the application wants to do. It doesn't precisely have to be a "client ID" either, but that term to me implies fine-grained context for "who" the caller is. If it makes sense to use the Passwordless API key for example, that works too.

It finally has some momentum to it, so over in Bitwarden Secrets Manager we are going to be investing in a similar way, using the context we have there. This won't be about the "client ID" necessarily but other parameters that come in via authorization headers, and we then have lots of capabilities to provide rate limiting.

Tyrrrz:
As discussed, the new approach works the following way:

- Rate limit bypass is set up on per-application basis, through the configuration, most likely populated by environment variables.
- It looks like this (in JSON form):
  ```jsonc
  {
	  "Serilog": {
	    "MinimumLevel": {
	      "Default": "Information",
	      "Override": {
	        "Microsoft": "Warning",
	        "Microsoft.Hosting.Lifetime": "Information",
	        "System": "Warning"
	      }
	    },
	    "Filter": [
	      {
	        "Name": "ByExcluding",
	        "Args": {
	          "expression": "Contains(RequestPath, '/health/') and StatusCode=200"
	        }
	      }
	    ]
	  },
	  "AllowedHosts": "*",
	  "SALT_TOKEN":  "",
	  "Datadog": {
	    "url": "https://http-intake.logs.datadoghq.eu"
	  },
	  "Fido2": {
	    "MDS": {
	      "Mode": "Online"
	    }
	  },
	  // This is the bit that matters
	  "ApplicationOverrides": {
	    "qa-app-1": {
	      "IsRateLimitBypassEnabled": true
	      // Room for other, similar overrides
	    }
	  }
	}
  ```
- Rate limit is then selectively disabled for the specified app(s)

Using environment variables, you can specify that override like so:

```
ApplicationOverrides__qa-app-1__IsRateLimitBypassEnabled=true
```

@abergs @jrmccannon

src/Api/Overrides/ApplicationOverrides.cs
```diff
@@ -0,0 +1,12 @@
...
+
+/// <summary>
+/// Configures behavior overrides for an application.
+/// </summary>
+public record ApplicationOverrides(string ApplicationId)
+{
+    /// <summary>
+    /// Whether actions on behalf of this application bypass rate limiting.
+    /// </summary>
+    public bool IsRateLimitBypassEnabled { get; init; }
```
Tyrrrz, line 11: This could also include `IsAdmin` for the admin console app

abergs (approved):
shape, security, code: Looks good

src/Api/Endpoints/Magic.cs
```diff
@@ -26,12 +27,20 @@ public static class MagicEndpoints
...
     public static void AddMagicRateLimiterPolicy(this RateLimiterOptions builder) =>
         builder.AddPolicy(RateLimiterPolicy, context =>
         {
             var tenant = context.User.FindFirstValue(CustomClaimTypes.AccountName) ?? "<global>";
 
+            var isRateLimitBypassed = context.RequestServices
+                .GetRequiredService<IConfiguration>()
+                .GetSection("ApplicationOverrides")
+                .TryGetApplicationOverrides(tenant)
+                ?.IsRateLimitBypassEnabled == true;
```
jrmccannon, line 39: Instead of a nullable, we could just have it return a default implementation that is false.

src/Api/Overrides/ApplicationOverridesExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    /// <summary>
+    /// Gets overrides for the specified application from the configuration.
+    /// </summary>
+    public static ApplicationOverrides? TryGetApplicationOverrides(
+        this IConfiguration configuration,
+        string applicationId) =>
+        configuration
+            .GetSection(applicationId)
+            .Get<ApplicationOverrides>();
```
jrmccannon, line 16: Could return default object instead of null

tests/Api.IntegrationTests/PasswordlessApiFixture.cs
```diff
@@ -11,10 +11,14 @@ public class PasswordlessApiFixture : IAsyncDisposable, IAsyncLifetime
     public async Task InitializeAsync() => await _dbContainer.StartAsync();
 
     public async Task<PasswordlessApi> CreateApiAsync(
-        ITestOutputHelper? testOutput = null,
-        bool disableRateLimiting = true)
+        IReadOnlyDictionary<string, string?> settings = null,
```
jrmccannon, line 14: Could this go second so you wouldn't have to update every method?

jrmccannon (commented):
Small comments

jrmccannon (approved):

PR #514, opened 2024-03-23
PAS-202 | OpenAPI improvements

jonashendrickx:
### Ticket
- Closes [PAS-202](https://bitwarden.atlassian.net/browse/PAS-202)

### Description

- Add response types for every endpoint
- Remove list of schemas at the bottom, which is now completely redundant given they're present on the endpoints themselves.
- Rename schema definition from v1 to v4.
- Apply CORS policy to the schema at /swagger/v4/swagger.json so we can use the schema with VuePress.
- Fix where `GET /authenticators/list` applies many ApiSecret headers to the schema.
- CSS cleanup

### Shape

n/a

### Screenshots

image

image

image

### Checklist
- __

- __

[PAS-202]: https://bitwarden.atlassian.net/browse/PAS-202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Api/Endpoints/Apps.cs
+++ b/src/Api/Endpoints/Apps.cs
@@ -1,4 +1,6 @@
 using System.ComponentModel.DataAnnotations;
+using System.Net;
+using System.Net.Mime;
 using Microsoft.AspNetCore.Mvc;
 using Passwordless.Api.Authorization;
 using Passwordless.Api.Helpers;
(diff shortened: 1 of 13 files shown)


src/Api/Endpoints/Apps.cs
```diff
@@ -222,6 +224,8 @@ public static async Task<IResult> IsAppIdAvailableAsync([AsParameters] GetAppIdA
     /// <summary>
     /// Change the configuration or features.
     /// </summary>
+    [ProducesResponseType((int)HttpStatusCode.NoContent)]
```
jrmccannon, line 227: Would we rather use the `StatusCodes` int constants instead of the enum values?

```
[ProducesResponseType(StatusCodes.Status204NoContent)]
```

jrmccannon (approved):

PR #534, opened 2024-04-03
PAS-406 | Check authenticators exist in MDS before configuration

jonashendrickx:
### Ticket
- Closes [PAS-406](https://bitwarden.atlassian.net/browse/PAS-406)

### Description

- Check if authenticator exists in MDS before configuration
- Show default label 'Authenticator' in admin console when a configured authenticator is removed in MDS since we do not own the data.

### Shape

n/a

### Screenshots

n/a

### Checklist
- __

- __

### DIFF
--- a/src/Service/MDS/IMetaDataService.cs
+++ b/src/Service/MDS/IMetaDataService.cs
@@ -7,4 +7,11 @@ public interface IMetaDataService
     Task<IEnumerable<string>> GetAttestationTypesAsync();
     Task<IEnumerable<EntryResponse>> GetEntriesAsync(EntriesRequest request);
     Task<IEnumerable<string>> GetCertificationStatusesAsync();
+
+    /// <summary>
+    /// Checks whether all the given authenticators exist in the FIDO2 MDS.
+    /// </summary>
+    /// <param name="aaGuids"></param>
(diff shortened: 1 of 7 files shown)


src/Service/MDS/IMetaDataService.cs
```diff
@@ -7,4 +7,11 @@ public interface IMetaDataService
     Task<IEnumerable<string>> GetAttestationTypesAsync();
     Task<IEnumerable<EntryResponse>> GetEntriesAsync(EntriesRequest request);
     Task<IEnumerable<string>> GetCertificationStatusesAsync();
+
+    /// <summary>
+    /// Checks whether all the given authenticators exist in the FIDO2 MDS.
+    /// </summary>
+    /// <param name="aaGuids"></param>
+    /// <returns>Returns `true` if all the authenticators exist in the FIDO2 MDS; otherwise, `false`.</returns>
+    Task<bool> IsExistsAsync(IReadOnlyCollection<Guid> aaGuids);
```
abergs, line 16: @jonashendrickx I think the naming here could just be `ExistsAsync`, as that seems to be what [dotnet uses](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.exists?view=net-8.0)?

jonashendrickx: @abergs  renamed

jrmccannon (approved):

PR #537, opened 2024-04-03
Upgrade Stripe SDK

jonashendrickx:
### Description

Upgrades the Stripe SDK

### Shape

We would have to monitor a production deployment, and upgrade the Stripe API version in the Stripe dashboard the exact moment the load balancer swaps the deployment slots.

### Checklist
- __

- __


jonashendrickx:
@abergs If you merge it now we get plenty of time to test.

abergs:
@jonashendrickx I don't think we need to upgrade the API version in the stripe dashboard if we update the SDK, are you sure about this?

jonashendrickx:
> @jonashendrickx I don't think we need to upgrade the API version in the stripe dashboard if we update the SDK, are you sure about this?

Yes. Setup the Stripe CLI and add a local listener. Then upgrade the Stripe SDK to verify. It will start throwing exceptions and return a 500.

PR #539, opened 2024-04-03
Refactoring `Admins`

jonashendrickx:
### Description

Continuing to hunt down table components, forms to get a more uniform styling in the entire application and further reduce technical debt.

### Shape

- There were quite a lot of redundant calls in the Razor Pages version which retrieved data more than once such as the organization, which was already being retrieved by middleware.
- You could invite again who was already an admin.

### Screenshots

image

### Checklist
- __

- __


src/AdminConsole/Components/Pages/Organization/Admins.razor.cs
```diff
@@ -0,0 +1,150 @@
...
+
+    private async Task InviteAsync()
+    {
+        if (CanInvite is false)
+        {
+            InviteAdminModelState = new ModelState("You need to upgrade to a paid organization to invite more admins.");
+            return;
+        }
+
+        if (ConsoleAdmins!.Any(x => string.Equals(x.Email, InviteForm.Email, StringComparison.OrdinalIgnoreCase)))
```
abergs, line 86: issue: This causes admin account enumeration. While the UX is not ideal, the way we solve this today is that we send an email informing the target email that they've been invited but can't join since an admin-email can only be connected to one organisation etc.

jrmccannon: I could be wrong, but `ConsoleAdmins` are shown on the Admins page and having this check wouldn't allow for additional account enumeration.

jonashendrickx: @abergs the only good way to fix it is by allowing the email to be sent. And then do the validation upon clicking the link. Then you avoid enumeration. And the one clicking the link owns the email.

abergs: Ah, I misunderstood. Thanks!

jrmccannon (approved):

PR #548, opened 2024-04-05
PAS-414 | Improve startup time by scheduling jobs to run at fixed intervals.

jonashendrickx:
### Ticket

- Closes [PAS-414](https://bitwarden.atlassian.net/browse/PAS-414)

### Description

We need to improve startup time. Some background jobs run immediately upon startup.

### Shape

- Ideally we should get rid of these sync jobs by improving the overall architecture and use a more reactive approach. We will otherwise run into scalability issues.
- user count updater is only used for admins currently and metered billing. We can migrate metered billing to use data from periodic credential reports which runs once a day. And then user count is otherwise updated when an admin logs in.
- Job runs once per hour, deploying within x seconds before the job is supposed to start could still result into a timeout. But is now much less likely to happen. We could still extend BasePeriodicBackgroundService with a built-in delay if the first run is to happen within 5 minutes to skip one cycle.

### Screenshots

image

### Checklist
- __

- __

[PAS-414]: https://bitwarden.atlassian.net/browse/PAS-414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


abergs:
I need some context to understand this change.

Is that screenshot before the change? What does does timestamp indicate?

jonashendrickx:
> I need some context to understand this change.
> 
> Is that screenshot before the change? What does does timestamp indicate?

It says when the service will run for the first time immediately after starting the Api or AdminConsole. It will also log when it starts stopping and when it has stopped completely. Before the change, the background services couldn't be interrupted or stopped as it wasn't listening to the cancellation token.

abergs (approved):
During we call, we identified that we do want to run these on startup. But we can opt for a delay of a minute or two.

abergs (dismissed):
(clicked the wrong button)

jonashendrickx:
> During we call, we identified that we do want to run these on startup. But we can opt for a delay of a minute or two.

Changes have been made. In a separate pull request later I will check if I can have metered billing take the data from the periodic credential reports. And then refresh the user counts as the admin logs in.

jrmccannon (approved):

PR #555, opened 2024-04-15
Fix some papercuts/issues when creating an org in dev mode

Tyrrrz:
1. Fixed an issue where the `Verify.cshtml` was referring to the wrong `mail.md` file (there is one for AdminConsole and one for API). Now there's only file, located at `../mail.md`. This is specified using configuration, not hard-codded.
2. `Verify.cshtml` now displayed the contents of `mail.md` so you don't have to manually look for it.
3. Removed `MagicClient` because that functionality is now part of `PasswordlessClient`.
4. Changed the order of admin credentials in app settings to match the order they're displayed during intiailization.
5. Made use of `async` overloads of some methods, where applicable.


src/AdminConsole/Services/Mail/DefaultMailService.cs
```diff
@@ -52,7 +52,7 @@ public async Task SendPasswordlessSignInAsync(string magicLink, string email)
         await _provider.SendAsync(message);
     }
```
jrmccannon, line 53: Github will only allow me to comment down here.

The method `SendPasswordlessSignInAsync` is unused now and can be deleted.

jrmccannon (changes requested):

src/AdminConsole/Services/Mail/DefaultMailService.cs
Tyrrrz: You can shift+click to select a region and comment on a code block

PR #569, opened 2024-04-22
PAS-438 | Touch up the readme with more information

Tyrrrz:


README.md
```diff
@@ -8,68 +8,147 @@ Bitwarden Passwordless.dev is a software toolkit that helps developers build FID
...
-    headers: {
-        "ApiSecret": "myapplication:secret:11f8dd7733744f2596f2a28544b5fbc4",
-        "Content-Type": "application/json"
+You can use Passwordless in conjunction with a variety of different backend platforms — see the [documentation](https://docs.passwordless.dev/guide/backend) for more info. Below is an example of a backend integration using ASP.NET Core and the [Passwordless SDK for .NET](https://github.com/bitwarden/passwordless-dotnet):
+
+```csharp
+// Add Passwordless to your service container
+services.AddPasswordlessSdk(options =>
+{
+    options.ApiSecret = "your_api_secret";
```
jonashendrickx, line 32: You don't need both. Just the secret.

jonashendrickx (changes requested):

README.md
Tyrrrz: The example in the docs sets up both... Should we update both?

jonashendrickx: I can't find it in the docs, but if it is there, then it has to be removed

README.md
```diff
@@ -8,68 +8,147 @@ Bitwarden Passwordless.dev is a software toolkit that helps developers build FID
...
+    options.ApiKey = "your_api_key";
```
Tyrrrz, line 33:
```suggestion
```

Tyrrrz:
@abergs @jonashendrickx any comments? 🙂

abergs (approved):

abergs:
@Tyrrrz Could we perhaps include a link to our demo or a gif on the UX? I think it's powerful.

Tyrrrz:
> @Tyrrrz Could we perhaps include a link to our demo or a gif on the UX? I think it's powerful.

Do we have such a demo/gif already?

Tyrrrz:
@abergs I added the demo video. There were two options on how to embed it into the readme:

1. GIF. This is the most straight-forward option and supports auto-play. However, the file size is quite big -- 11mb. I can sacrifice the quality or the framerate to compress it, but then it looks quite ugly.
2. Video element. It doesn't support auto-play. But it allows us to embed a video file, namely `webm` in this case, which is only 800kb in size without any loss in quality. The file is also not part of the repository but instead hosted by GitHub, so it won't affect cloning.

I went with option 2. This is what it looks like: https://github.com/bitwarden/passwordless-server/tree/Tyrrrz-patch-1?tab=readme-ov-file#demo

Let me know if this is good enough.

abergs:
Solid! @Tyrrrz

abergs (approved):

jonashendrickx (approved):

PR #576, opened 2024-05-01
PAS-444 | Populate metered billing from periodic credential reports.

jonashendrickx:
### Ticket

- Closes [PAS-444](https://bitwarden.atlassian.net/browse/PAS-444)

### Description

This is the first step to populate metered billing from periodic credential reports in the api. Stripe recommends we keep the usage records ourselves, so for this the periodic credential reports will serve their purpose.

In step 2 of the refactoring, we will stop updating the CurrentUserCount in the Applications table periodically, but instead do it upon login of an admin to significantly reduce network traffic.

### Screenshots

For testing I populated PeriodicCredentialReports first. Then forced the background job to run:

image

### Checklist
- Tested locally.

- __

[PAS-444]: https://bitwarden.atlassian.net/browse/PAS-444?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/AdminConsole/Billing/BackgroundServices/MeteredBillingBackgroundService.cs
+++ b/src/AdminConsole/Billing/BackgroundServices/MeteredBillingBackgroundService.cs
@@ -1,5 +1,9 @@
+using Passwordless.AdminConsole.Db;
+using Passwordless.AdminConsole.Middleware;
 using Passwordless.AdminConsole.Services;
 using Passwordless.Common.Background;
+using Passwordless.Common.Models.Reporting;
+using Stripe;
 
 namespace Passwordless.AdminConsole.Billing.BackgroundServices;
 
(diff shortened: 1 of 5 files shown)


src/AdminConsole/Billing/BackgroundServices/MeteredBillingBackgroundService.cs
```diff
@@ -17,8 +21,71 @@ protected override async Task DoWorkAsync(CancellationToken cancellationToken)
...
+            Logger.LogInformation("Updating metered billing for {Organizations} organizations.", organizations.Count);
+            foreach (var organization in organizations)
+            {
+                var items = new Dictionary<string, int>();
+
+                foreach (var application in organization.Applications)
+                {
+                    Logger.LogInformation("Processing metered billing for application '{Application}'.", application.Id);
+
+                    var context = scope.ServiceProvider.GetRequiredService<ICurrentContext>();
```
jonashendrickx, line 41: Set the `ICurrentContext` so the `ApiSecret` is being set properly.

src/AdminConsole/Billing/BackgroundServices/MeteredBillingBackgroundService.cs
```diff
@@ -17,8 +21,71 @@ protected override async Task DoWorkAsync(CancellationToken cancellationToken)
...
+                    context.SetApp(application);
+
+                    var request = new PeriodicCredentialReportRequest(DateOnly.FromDateTime(timeProvider.GetUtcNow().Date.AddDays(-1)), null);
+                    var api = scope.ServiceProvider.GetRequiredService<IScopedPasswordlessClient>();
```
jonashendrickx, line 45: We initialize the `IScopedPasswordlessClient` implementation here which will use the correct `ApiSecret` to retrieve the metered billing information.

jrmccannon (approved):
This is good for a first pass.

I worry the double loop to the API doesn't scale real well, but the next step will alleviate some of this.

Will this also run on multiple instances of AdminConsole?

PR #583, opened 2024-05-12
PAS-446 | Fix memory leak in Serilog

jonashendrickx:
### Ticket

- Closes [PAS-446](https://bitwarden.atlassian.net/browse/PAS-446)

### Description

Fix memory leak caused by Serilog.

### Checklist
- Load tested locally for several hours.

- __

[PAS-446]: https://bitwarden.atlassian.net/browse/PAS-446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Api/Program.cs
```diff
@@ -95,11 +90,7 @@
                     configuration: ddConfig);
             }
         }
-    },
-    false,
-    // Pass log events down to other logging providers (e.g. Microsoft) after Serilog, so
-    // that they can be processed in a uniform way in tests.
-    true
```
abergs, line 102: @jonashendrickx issue: This was added for a reason, if we just remove it we'll break logging in tests as indicated through the comment/git history.

My suggestion was that we set this argument depending on if we're running in env.Prod instead - was there an issue with my suggestion?

jonashendrickx: I just couldn't read the suggestion in the chat - the thread refused to load.

abergs (commented):
Looks good - how can we test that the logs are getting passed through during CI/CD? (cc @Tyrrrz)

abergs:
LGTM, but were we able to test it?

jonashendrickx:
> LGTM, but were we able to test it?

The tests passed. I would be surprised the environment variable for that specific step wouldn't work. We might just have to create a separate flag for it later if running tests in Development mode conflicts with something else.

Tyrrrz:
> Looks good - how can we test that the logs are getting passed through during CI/CD? (cc @Tyrrrz)

It seems that they're there

image

PR #591, opened 2024-05-17
Configure admin console overrides in self-host

Tyrrrz:


Tyrrrz:
Note I still can't seem to test it locally because running the locally built container results in:

```console
exec /entrypoint.sh: no such file or directory
```

For whatever reason

jonashendrickx (approved):

PR #600, opened 2024-05-31
PAS-462 | BEEEP Migrate `Access Denied` to Blazor.

jonashendrickx:
### Ticket

- Closes [PAS-462](https://bitwarden.atlassian.net/browse/PAS-462)

### Description

- Migrates page to Blazor.
- Fixes spelling.

### Shape

n/a

### Checklist
- Tested locally

- __

[PAS-462]: https://bitwarden.atlassian.net/browse/PAS-462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Common/Extensions/UrlExtensions.cs
```diff
@@ -0,0 +1,65 @@
+namespace Passwordless.Common.Extensions;
+
+public static class UrlExtensions
+{
+    public static bool IsLocalUrl(this string? url)
```
Tyrrrz, line 5: Why not use the `Uri` class?

src/Common/Extensions/UrlExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        if (string.IsNullOrEmpty(url))
+        {
+            return false;
+        }
```
Tyrrrz, line 10: Extension methods typically shouldn't be callable on `null` instances (similar to member methods) to avoid confusion

src/Common/Extensions/UrlExtensions.cs
jonashendrickx: This is literally the code we were using in Razor Pages.

```https://github.com/dotnet/aspnetcore/blob/main/src/Mvc/Mvc.Core/src/Routing/UrlHelperBase.cs```

PR #617, opened 2024-06-14
Added auth configs to playground

jrmccannon:
### Ticket
- Closes [PAS-397](https://bitwarden.atlassian.net/browse/PAS-397)

### Description
Added authentication configurations to the playground sign in form.

### Shape
In order to facilitate easier testing of using step up method, I upgraded the passwordless client-js package to 1.2.0-beta1. It contains the stepup method and changes to `signinBegin`.

This allows for swapping between the different configurations and validating that users can successfully retrieve a valid token for each authentication.

### Screenshots
Sign In:
image

Step Up Hidden Options:
image

Step Up Shown Options:
image

### Checklist
- sign-in works fine
- step-up works fine
- custom auth configs work as well
- regular authentication to adminconsole works (since we ugraded the client-js version)

- n/a

[PAS-397]: https://bitwarden.atlassian.net/browse/PAS-397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jonashendrickx:
I would personally add a tab to each playground page so we can split it up in the following:

- registration
- sign-in
- step up

The step up page can then contain a bit more elements for the user to sign into a specific context, and then obtain a second context, to properly demonstrate what happens. Then we can possibly also avoid this page from potentially becoming too large.

My first impression was that it might be `too crowded` now with too many options, which might then turn people away and prevent them from getting started because there's just too much information at once to process.

src/AdminConsole/Pages/App/Playground/Client.cshtml
```diff
@@ -27,6 +27,11 @@
             <div class="mt-8">
                 <div class="mt-6">
                     <form v-on:submit.prevent="loginSubmit" href="#" name="signin" method="post">
+                        <div class="hidden sm:block">
+                            <div class="border-b border-gray-200 -mb-px flex" aria-label="Tabs">
+                                <button v-for="tab in tabs" :key="tab.name" v-on:click.prevent="setTab(tab)" :class="[tab.current ? 'border-indigo-500 text-indigo-600' : 'border-transparent text-gray-500 hover:border-gray-300 hover:text-gray-700', 'flex-grow border-b-2 px-1 py-4 text-center text-sm font-medium']" :aria-current="tab.current ? 'page' : undefined">{{ tab.name }}</button>
+                            </div>
+                        </div>
                         <fieldset id="registerCredentialFields" class="space-y-6">
```
Tyrrrz, line 35: Can you add some top margin here so the text appears further away from the tab strip?

src/AdminConsole/Pages/App/Playground/Client.cshtml
```diff
@@ -27,6 +27,11 @@
             <div class="mt-8">
                 <div class="mt-6">
                     <form v-on:submit.prevent="loginSubmit" href="#" name="signin" method="post">
+                        <div class="hidden sm:block">
```
Tyrrrz, line 30: Maybe show a dropdown on mobile so as to offer something beyond `hidden`?

src/AdminConsole/Pages/App/Playground/Client.cshtml
```diff
@@ -141,10 +192,17 @@
                     signins.value.push(result);
                     console.log(signins.value);
                 }
-
+
```
Tyrrrz, line 195:
```suggestion

```

src/AdminConsole/Pages/App/Playground/Client.cshtml
Tyrrrz: Also, since the purposes are dynamic and can theoretically be a very long list, maybe showing a drop down on desktop is also desirable.

Tyrrrz (approved):

PR #630, opened 2024-07-04
PAS-386 | Don't require restarting the app during Admin Console initialization

Tyrrrz:
### Ticket

- Closes https://bitwarden.atlassian.net/browse/PAS-386

### Description

Currently, Admin Console asks the user to restart the app after completing the initialization. This PR makes it unnecessary by tuning the order in which the services/options are initialized and making use of the `reloadOnChange` feature in `Microsoft.Extensions.Configuration`.

### Shape

- Don't initialize the Passwordless SDK if the credentials are missing. Previously it was initialized with fake, non-working credentials as a work-around.
- Don't require the Passwordless client in code paths used during initialization. This allows the Admin Console to perform initialization even if the `Passwordless:` section is not configured.
- Fetch the Passwordless options in `SetupService` in a way that's not dependent on the initialization of the Passwordless SDK. The SDK does not allow the options to contain a `null` or empty `ApiSecret`. In `SetupService` this is a valid scenario, which indicates that the initialization has not been completed yet.
- Update the text on the page to instruct the user what to do after updating the configuration file.

### Screenshots

Step by step:

2024-07-04 18_36_01-localhost_8001_initialize

2024-07-04 18_37_47-localhost_8001_initialize

2024-07-04 18_38_02-PasswordlessServer – appsettings Development json

2024-07-04 18_38_12-Home page - Admin Console - Bitwarden Passwordless dev

### Checklist

- Ran the initialization flow locally

- Did not find any new vectors

### DIFF
--- a/src/AdminConsole/Services/MagicLinks/MagicLinkBuilder.cs
+++ b/src/AdminConsole/Services/MagicLinks/MagicLinkBuilder.cs
@@ -9,7 +9,7 @@ namespace Passwordless.AdminConsole.Services.MagicLinks;
 
 public class MagicLinkBuilder(
     SignInManager<ConsoleAdmin> signInManager,
-    IPasswordlessClient passwordlessClient,
+    IServiceProvider services,
     IActionContextAccessor actionContextAccessor,
     IUrlHelperFactory urlHelperFactory,
     ILogger<MagicLinkBuilder> logger)
@@ -17,6 +17,11 @@ public class MagicLinkBuilder(
 {
     public async Task<string> GetLinkAsync(string email, string? returnUrl = null)
     {
+        // Passwordless client is fetched lazily because the MagicLinkBuilder may need to instantiate
+        // in a context where the API is not accessible, for example when the Admin Console has not
+        // been initialized yet. Doing this prevents it from crashing unless the client is actually used.
+        var passwordlessClient = services.GetRequiredService<IPasswordlessClient>();
+
         var user = await signInManager.UserManager.FindByEmailAsync(email);
         if (user == null)
         {
@@ -25,7 +30,12 @@ public class MagicLinkBuilder(
         }
 
         var token = await passwordlessClient.GenerateAuthenticationTokenAsync(new AuthenticationOptions(user.Id));
-        var urlBuilder = urlHelperFactory.GetUrlHelper(actionContextAccessor.ActionContext ?? throw new InvalidOperationException("ActionContext is null"));
+
+        var urlBuilder = urlHelperFactory.GetUrlHelper(
+            actionContextAccessor.ActionContext ??
+            throw new InvalidOperationException("ActionContext is null")
+        );
+
         var url = urlBuilder.PageLink("/Account/Magic", values: new { returnUrl, token }) ?? urlBuilder.Content("~/");
 
         return url;
@@ -33,9 +43,13 @@ public class MagicLinkBuilder(
 
     public string GetUrlTemplate(string? returnUrl = null)
     {
-        var urlBuilder = urlHelperFactory.GetUrlHelper(actionContextAccessor.ActionContext ?? throw new InvalidOperationException("ActionContext is null"));
+        var urlBuilder = urlHelperFactory.GetUrlHelper(
+            actionContextAccessor.ActionContext ??
+            throw new InvalidOperationException("ActionContext is null")
+        );
 
-        var url = urlBuilder.PageLink("/Account/Magic", values: new { token = "$TOKEN", returnUrl }) ?? urlBuilder.Content("~/");
+        var url = urlBuilder.PageLink("/Account/Magic", values: new { token = "$TOKEN", returnUrl }) ??
+                  urlBuilder.Content("~/");
 
         return url;
     }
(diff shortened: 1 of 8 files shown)


src/AdminConsole/Services/MagicLinks/MagicLinkBuilder.cs
```diff
@@ -9,14 +9,19 @@ namespace Passwordless.AdminConsole.Services.MagicLinks;
...
     IUrlHelperFactory urlHelperFactory,
     ILogger<MagicLinkBuilder> logger)
     : IMagicLinkBuilder
 {
     public async Task<string> GetLinkAsync(string email, string? returnUrl = null)
     {
+        // Passwordless client is fetched lazily because the MagicLinkBuilder may need to instantiate
+        // in a context where the API is not accessible, for example when the Admin Console has not
+        // been initialized yet. Doing this prevents it from crashing unless the client is actually used.
+        var passwordlessClient = services.GetRequiredService<IPasswordlessClient>();
```
Tyrrrz, line 23: This is not clean but this was the easiest solution.

A better solution would be to provide some sort of `IPasswordlessClientProxy` that would implement the `IPasswordlessClient` contract, but would instead initialize the client lazily each time any method is called. I think if we do go this route, this proxy should be provided by the .NET SDK directly to avoid duplicating the client initialization logic, especially when it comes to Identity integration.

Let me know what you think.

jonashendrickx (commented):
Pull request title seems missleading, I can mostly only see code formatting?

Tyrrrz:
@jonashendrickx for functional changes, see `Initialize.razor`, `MagicLinkBuilder.cs` and `SetupService.cs`

src/AdminConsole/Services/SetupService.cs
```diff
@@ -1,18 +1,17 @@
...
+    private readonly PasswordlessOptions? _options;
     private readonly bool _isSelfHosted;
 
     public SetupService(
-        IOptions<PasswordlessOptions> options,
         IConfiguration configuration)
     {
-        _options = options.Value;
+        // Get the Passwordless options manually to skip the built-in validations in the Passwordless SDK
+        _options = configuration.GetSection("Passwordless").Get<PasswordlessOptions>();
```
jonashendrickx, line 14: Use IOptionsSnapshot.

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-8.0#use-ioptionssnapshot-to-read-updated-data

jonashendrickx (changes requested):

src/AdminConsole/Services/SetupService.cs
Tyrrrz: It's not possible to use `IOptionsSnapshot` for the reason explained in the comment.

jonashendrickx: @Tyrrrz There is nothing wrong with the order of the configurations. I can help you with it next week.

Line 79 in:
```src/AdminConsole/Db/DatabaseBootstrap.cs```

Take a look at how it's configured. The lambda expression doesn't support `reloadOnChange`.

Change it to `.AddPasswordless("Passwordless")` and it would work just fine both in Docker and using the SelfHosting run configuration.

However doing so would break mapping the `InternalApiUrl`. So we'd probably have to fix this.

See also:
- https://github.com/bitwarden/passwordless-dotnet/pull/124
- https://github.com/bitwarden/passwordless-dotnet/pull/125

`InternalApiUrl` under `PasswordlessManagement` was meant to prevent network traffic to unnecessarily leave the container.

Tyrrrz: @jonashendrickx I think you're a bit confused in regard to what is the actual problem here. Reloading works completely fine. `IOptionsSnapshot<>` would work fine too, if not for how `AddPasswordless()` works.

The core issue is here:

https://github.com/bitwarden/passwordless-dotnet/blob/aa04a1d7a981b9b998e1c45e20c32937c5aab983/src/Passwordless/ServiceCollectionExtensions.cs#L22-L25

As you can see, the options are created with a validator that checks that the `ApiSecret` is not null or empty. In other words, when we call `AddPasswordless`, the secret must already have been set.

However, during the initialization sequence, it is not set yet. The whole purpose of the initialization sequence is to generate the `ApiSecret` and add it to the configuration. So in order to check whether it has been added yet or not, we can't rely on the `PasswordlessOptions` provided by the Passwordless SDK since they would fail to initialize immediately, seeing as `ApiSecret` is null.

jonashendrickx: Seems like whatever solution we would go with it would be personal preference.

Tyrrrz:
Will rework without relying on current hacks to skip validation in the SDK

Tyrrrz:
Reworked by manually re-registering `PasswordlessAspNetCoreOptions` from `IConfiguration`. Can be further improved by making changes in the Passwordless SDK but I decided to do it here to unblock this PR and get it merged. After making changes in the SDK, this can be polished quite trivially.

Tyrrrz:
Updated the SDK, reworked this PR and now it doesn't have any hacks

abergs (approved):
Shape: ok!

PR #636, opened 2024-07-13
PAS-452 | Design generic fallback for e-mail providers & drop PostMark

jonashendrickx:
### Ticket

- Closes [PAS-452](https://bitwarden.atlassian.net/browse/PAS-452)

### Description

This pull request removes support for Postmark.

Any e-mail provider, including Postmark, can still continue to be configured using SMTP. Read more [here](https://docs.passwordless.dev/guide/self-hosting/configuration.html#e-mail).

### Shape

Features:
- Allow configuration of an ordered list of mail providers.
- Can combine multiple mail providers of the same type, i.e. AWS. For example, can be used if one region is down, and fall back to using a different region. This is purely an example, as not every mail provider could support this out of the box.
- When no mail provider is configured, use the `FileMailProvider`
- Allow updating the provider configurations without having to restart the application.

In the example below, we're able to attempt to send e-mails first using AWS SES using region us-west-2, if this fails, we fall back to AWS SES using region us-east-2. If that again fails, we attempt to send e-mails using SendGrid. Please note the example is purely hypothetical and ignores any vendor specific features such as the availability zones with redundant connections with AWS SES to guarantee minimum down-time.

If all configured attempts fail, an error will be thrown by the back-end.

```
  "Mail": {
    "From": "johndoe@example.com",
    "Providers": [
      {
        "Name": "aws",
        "AccessKey": "aws_access_key_id",
        "SecretKey": "aws_secret_key",
        "Region": "us-west-2"
      },
      {
        "Name": "aws",
        "AccessKey": "aws_access_key_id",
        "SecretKey": "aws_secret_key",
        "Region": "us-east-1"
      },
      {
        "Name": "sendgrid",
        "ApiKey": "sendgrid_api_key"
      }
    ]
  },
```

### Checklist
- __

- __

[PAS-452]: https://bitwarden.atlassian.net/browse/PAS-452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


abergs:
@jonashendrickx we're not merging this until more work is done in PAS-452, right?

src/Common/Services/Mail/Aws/AwsMailProviderOptions.cs
```diff
@@ -0,0 +1,12 @@
+namespace Passwordless.Common.Services.Mail.Aws;
+
+public class AwsMailProviderOptions : BaseMailProviderOptions
+{
+    public const string Provider = "aws";
```
Tyrrrz, line 5: `base.Name` does not appear to be set in the constructor, like in `FileMailProviderOptions`?

src/Common/Services/Mail/MailBootstrap.cs
```diff
@@ -1,39 +1,55 @@
...
 
-            var clients = new List<PostmarkClientConfiguration>();
-            configurationSection.GetSection("MessageStreams").Bind(clients);
+                // Add a default file provider if no providers are configured.
+                if (!section.GetChildren().Any())
+                {
+                    var fileProviderOptions = new FileMailProviderOptions();
+                    o.Providers.Add(fileProviderOptions);
+                    return;
+                }
```
Tyrrrz, line 27: I think we should have a no-op mail provider in the default case. This may be useful in tests, where you don't care about emails at all and the app in the container may not have the permissions to have files.

src/Common/Services/Mail/MailConfiguration.cs
```diff
@@ -2,5 +2,13 @@ namespace Passwordless.Common.Services.Mail;
...
 {
+    /// <summary>
+    /// The default email address to use as the sender.
+    /// </summary>
     public string? From { get; set; }
+
+    /// <summary>
+    /// The ordered list of mail providers to use.
+    /// </summary>
+    public List<BaseMailProviderOptions> Providers { get; set; } = new();
```
Tyrrrz, line 13: I think it should be `IReadOnlyList<>` to avoid accidental mutations after the configuration has been initialized. In `AddMail()` you can do: `var providers = new List<...>();` and then `o.Providers = providers`.

src/Common/Services/Mail/MailProviderFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public IMailProvider Create(string name, BaseMailProviderOptions options)
+    {
+        switch (name)
+        {
+            case AwsMailProviderOptions.Provider:
+                var awsOptions = (AwsMailProviderOptions)options;
+                var awsMailProviderLogger = _serviceProvider.GetRequiredService<ILogger<AwsMailProvider>>();
+                return new AwsMailProvider(awsOptions, awsMailProviderLogger);
```
Tyrrrz, line 24: Why not register the providers as services and let `IServiceProvider` handle dependencies?

src/Common/Services/Mail/OrderedMailProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+using Passwordless.Common.Services.Mail.File;
+using Passwordless.Common.Services.Mail.SendGrid;
+using Passwordless.Common.Services.Mail.Smtp;
+
+namespace Passwordless.Common.Services.Mail;
+
+/// <summary>
+/// Wraps multiple mail providers and tries to send the message using them in order.
+/// </summary>
+public class OrderedMailProvider : IMailProvider
```
Tyrrrz, line 12: Maybe `AggregateMailProvider` is a better name?

src/Common/Services/Mail/MailBootstrap.cs
jonashendrickx: We can add support for it later in the integration testing project

Tyrrrz (approved):

PR #637, opened 2024-07-16
PAS-47 | Allow admins to disable magic links for all admins belonging to an organization.

jonashendrickx:
### Ticket

- Closes [PAS-47](https://bitwarden.atlassian.net/browse/PAS-47)

### Description

This pull request allows an admin to disable or enable magic links for an entire organization.

### Shape

- Migrated old `/Organization/Settings` to Blazor.
- Each section in `/Organization/Settings` is isolated in its own Razor/Blazor component.
- During login, when the user is found, we check its organization whether the magic links flag is enabled, and handle accordingly in the page handler whether to display an error message or continue sending a magic link.

- Limitation: If we had an application per organization to hold credentials for admins, it would have been easier to implement, as we can just listen for the response of sending out magic links and display the error message returned.

### Screenshots

image

image

image

### Checklist
- Tried logging in with magic links disabled for a given organization.
- Tried logging in with magic links enabled for a given organization.
- Tried logging in with passkeys while magic links enabled for a given organization.
- Tried logging in with passkeys while magic links disabled for a given organization.
- Attempted to disable magic links in organization settings.
- Attempted to enable magic links in organization settings.
- Attempted to delete an organization in organization settings.

- __

[PAS-47]: https://bitwarden.atlassian.net/browse/PAS-47?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/AdminConsole/Components/Pages/Organization/SettingsComponents/DeleteOrganizationComponent.razor.cs
```diff
@@ -0,0 +1,91 @@
...
+
+    protected override void OnInitialized()
+    {
+        DeleteFormEditContext = new EditContext(DeleteForm);
+        DeleteFormValidationMessageStore = new ValidationMessageStore(DeleteFormEditContext);
+    }
+
+    private async Task OnDeleteFormSubmittedAsync()
+    {
+        if (!CanDelete)
```
jonashendrickx, line 43: We did not have this check here, could have been a backdoor to delete an organization by posting the form directly.

abergs:
Holding this to work out the complexity of the PR.

src/AdminConsole/Pages/Organization/Settings.cshtml
```diff
@@ -1,32 +0,0 @@
-﻿@page
-@model SettingsModel
-
-@{
-    ViewData["Title"] = "Organization Settings";
-}
-
-<div id="delete-organization" class="bg-white shadow sm:rounded-lg flex flex-col space-y-4 px-4 py-5 sm:p-6">
```
jonashendrickx, line 8: This section gets migrated to `DeleteOrganizationComponent.razor` without additional modifications.

src/AdminConsole/Components/Pages/Organization/Settings.razor
```diff
@@ -0,0 +1,16 @@
+@page "/Organization/Settings"
+@using Passwordless.AdminConsole.Components.Pages.Organization.SettingsComponents
+
+@inject IDataService DataService
+
+<Page Title="Organization Settings">
+    @if (_organization != null)
+    {
+        <SecurityComponent
```
jonashendrickx, line 9: Will hold all the different sections.

src/AdminConsole/Components/Pages/Organization/SettingsComponents/DeleteOrganizationComponent.razor.cs
```diff
@@ -0,0 +1,91 @@
...
+
+    public ValidationMessageStore? DeleteFormValidationMessageStore { get; set; }
+
+    /// <summary>
+    /// Whether the organization can be deleted.
+    /// </summary>
+    public bool CanDelete => ApplicationsCount == 0;
+
+    protected override void OnInitialized()
+    {
```
jonashendrickx, line 36: Unlike in Razor Pages, we can now pass down the minimum amount of data we need.

src/AdminConsole/Db/ConsoleDbContext.cs
```diff
@@ -41,6 +41,10 @@ protected override void OnModelCreating(ModelBuilder builder)
         builder.Entity<OrganizationEvent>()
             .HasKey(x => x.Id);
 
+        builder.Entity<Organization>()
```
jonashendrickx, line 44: We will store whether magic links are enabled on the `Organization` entity, and enable it by default. This should give us also backwards compatibility without issues.

src/AdminConsole/Pages/Account/Login.cshtml.cs
```diff
@@ -23,7 +29,8 @@ public IActionResult OnGet(string? returnUrl = null)
             return LocalRedirect(returnUrl);
         }
 
-        EmailSent = TempData["EmailSent"] is not null;
```
jonashendrickx, line 26: We replace this with an enumerable, so we can store the actual status and then display a message based of the current status in that temporary dictionary.

src/AdminConsole/Pages/Account/Login.cshtml.cs
```diff
@@ -32,16 +39,34 @@ public async Task<IActionResult> OnPost(string email, string? returnUrl = null)
...
         returnUrl = Url.Page("/Organization/Overview");
 
+        var user = await _signInManager.UserManager.FindByEmailAsync(email);
+
+        if (user == null)
+        {
+            return RedirectToPage();
+        }
+
+        var organization = await _dataService.GetOrganizationAsync(user.OrganizationId);
```
jonashendrickx, line 49: 1. If user is not found, abort and refresh the page.
2. If organization is not found related to the user that was found, abort and refresh the page (would never happen).
3. If magic links is disabled, set the correct status and refresh the page so we can see the error message.

src/AdminConsole/Services/AdminService.cs
```diff
@@ -0,0 +1,52 @@
...
+        IHttpContextAccessor httpContextAccessor,
+        ILogger<AdminService> logger)
+    {
+        _passwordlessClient = passwordlessClient;
+        _db = db;
+        _logger = logger;
+        _organizationId = httpContextAccessor.HttpContext!.User.GetOrgId()!.Value;
+    }
+
+    public async Task<bool> CanDisableMagicLinksAsync()
```
jonashendrickx, line 26: Checks whether there are admins without credentials. We could iterate over the `adminconsole` app and list all credentials, but in our hosted environment this could potentially return over >10k records. So this is a lower effort alternative without adding extra endpoints.

src/AdminConsole/Components/Pages/Organization/SettingsComponents/SecurityComponent.razor
```diff
@@ -0,0 +1,25 @@
+@inject IDataService DataService
+@inject IAdminService AdminService
+
+<Panel Header="Security">
+    @if (SecurityForm != null)
+    {
+    <EditForm class="flex flex-col space-y-4" FormName="@SecurityFormName" Model="SecurityForm" OnValidSubmit="@OnSecurityFormSubmittedAsync">
+        <div class="flex flex-col space-y-4">
+            <div class="flex flex-col space-y-2">
```
jrmccannon, line 9: We should probably have some wording here about what exactly "Enable Magic Links" Is.

If an admin loses their passkey, they will be locked out of the AdminConsole if this is disabled.  This should be made explicitly clear.

src/AdminConsole/Pages/Account/Login.cshtml.cs
```diff
@@ -32,16 +39,34 @@ public async Task<IActionResult> OnPost(string email, string? returnUrl = null)
     {
         returnUrl = Url.Page("/Organization/Overview");
 
+        var user = await _signInManager.UserManager.FindByEmailAsync(email);
+
+        if (user == null)
+        {
+            return RedirectToPage();
```
jrmccannon, line 46: This allows for user enumeration.

When we don't find a user currently, we still show that an email has been sent to the address provided.

If we want to protect against that, we should probably respond similarly to a successful submission.

src/AdminConsole/Pages/Account/Login.cshtml.cs
jrmccannon: Step 3 also allows for account enumeration by identifying users that belong to orgs that have disabled magic links.

One suggestion would be to still send an email, but just have the email not be a magic link email.

Instead it would be informing the user that their org has disabled magic link sign ins and to contact their admin, a message similar to the one implemented on the page.

src/AdminConsole/Components/Pages/Organization/SettingsComponents/SecurityComponent.razor.cs
```diff
@@ -0,0 +1,50 @@
...
+            if (!canDisableMagicLinks)
+            {
+                SecurityFormValidationMessageStore!.Add(() => SecurityForm.IsMagicLinksEnabled, "Cannot disable magic links because there are admins without passkeys.");
+                SecurityFormEditContext!.NotifyValidationStateChanged();
+                SecurityForm.IsMagicLinksEnabled = IsMagicLinksEnabled;
+                return;
+            }
+        }
+
+        await DataService.UpdateOrganizationSecurityAsync(SecurityForm!.IsMagicLinksEnabled);
```
jrmccannon, line 42: When running this locally, I cannot get Is magic links to re-enable.

jrmccannon (changes requested):
Replacement of the Org settings page with Blazor looks fine other that Enabling magic links not working to enable the setting. 

We should consider if we want better messaging on the login page directly or provide the feedback to the user in some other manner.  If we want the improved messaging directly on the page, this allows for account enumeration.

src/AdminConsole/Pages/Account/Login.cshtml.cs
jonashendrickx: @jrmccannon Check now, I am showing the same message as when magic links are sent. And it's sending the e-mail.

src/AdminConsole/Components/Pages/Organization/SettingsComponents/SecurityComponent.razor
jonashendrickx: image

jonashendrickx: Magic links allow administrators to sign into the admin console without a password.
- When enabled, if an administrator does not have his or her passkey, they can sign in with a magic link sent to their email. However, if an administrator's email is compromised, the attacker can sign in as the administrator.
- When disabled, administrators must sign in with their passkeys. If their passkeys are lost, a different administrator must reset their passkeys.

PR #650, opened 2024-07-25
PAS-540 | Parameters were being incorrectly passed

jonashendrickx:
### Ticket

- Closes [PAS-540](https://bitwarden.atlassian.net/browse/PAS-540)

### Description

E-mails were not being sent when magic links were disabled because the parameters were not correctly mapped. This wasn't caught by the `FileMailProvider`.

### Checklist
- __

- __

[PAS-540]: https://bitwarden.atlassian.net/browse/PAS-540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Common/Validation/EmailAddressValidator.cs
```diff
@@ -0,0 +1,51 @@
+using System.Globalization;
+using System.Text.RegularExpressions;
+
+namespace Passwordless.Common.Validation;
+
+public static class EmailAddressValidator
```
jrmccannon, line 6: We could just use `MailAddress` from `System.Net.Mail` instead of maintaining our own validation for email addresses.

src/AdminConsole/Services/Mail/DefaultMailService.cs
```diff
@@ -62,12 +63,17 @@ public async Task SendEmailIsAlreadyInUseAsync(string email)
 
     public Task SendMagicLinksDisabledAsync(string organizationName, string email)
     {
+        if (!EmailAddressValidator.IsValid(email))
```
jrmccannon, line 66: Changing the method signature to take a `System.Net.Mail.MailAddress` instead of a `string` would make the method easier to call correctly and give us the added benefit of built in validation of the email.

jrmccannon (changes requested):

jrmccannon (approved):

PR #669, opened 2024-08-13
Check for secrets before using them

withinfocus:
### Ticket

Internal change.

### Description

Adds conditionals to check for secret existence in workflows so they don't fail on forks.


withinfocus:
Test failure is transient.

jonashendrickx (approved):

PR #689, opened 2024-08-20
Improve integration tests performance

jonashendrickx:
### Description

Significantly reduces the resources required to run the integration tests as it reuses the Microsoft SQL Server test container.

If the tests are properly written, we should not see any issues. In case we do need a unique instance, we can just use a different collection identifier.

### Checklist
- __

- __


Tyrrrz (approved):

Tyrrrz:
> In case we do need a unique instance, we can just use a different collection identifier.

We can also re-use the same MSSQL instance but access different database instances.

PR #695, opened 2024-08-22
Add hacky validation for hints during registration

Tyrrrz:
Part of PAS-472

Since we're sending the form data using Vue and completely bypassing Razor Pages request pipeline, the validation attributes are not properly honored. I've added validation for Hints in Vue to avoid vague 400 errors on invalid hints. This is hacky, but a proper solution would require rewriting the page to not use Vue to send the registration request, or have better validation infrastructure on our front-end. I don't think it's worth the effort, given other priorities.


jonashendrickx:
Without rewriting the razor page, you could just define the page handler as an endpoint, and add the validation on the minimal api endpoint similarly to what we do in the back-end.

jrmccannon (approved):

PR #714, opened 2024-08-30
When no mail providers are registered, fall back to no-op behavior

Tyrrrz:
#705 changed the behavior in regard to how the mail providers are configured:
- Pre-PR, when the application configuration does not specify any mail provider, the emails are not sent.
- Post-PR, when the application configuration does not specify any mail provider, the `FileMailProvider` is used.

The issue is that, in certain scenarios, the application may not be able to write data to the file system and thus may not use the `FileMailProvider`, making it a poor default choice. This happens, for example, when running in Docker, where the application is executed under an under-privileged user account for security reasons:

image

Additionally, there does not appear to be a way to set the `NoopMailProvider` through configuration as it's only defined in tests.

This PR reverts some of the user-facing behavior changes introduced in #705 and makes it so that no emails are sent when the configuration does not specify any mail provider.

### DIFF
--- a/src/Common/Services/Mail/AggregateMailProvider.cs
+++ b/src/Common/Services/Mail/AggregateMailProvider.cs
@@ -5,53 +5,32 @@ namespace Passwordless.Common.Services.Mail;
 /// <summary>
 /// Wraps multiple mail providers and tries to send the message using them in order.
 /// </summary>
-public class AggregateMailProvider : IMailProvider
+public class AggregateMailProvider(
+    IOptionsSnapshot<MailConfiguration> options,
+    IMailProviderFactory factory,
+    ILogger<AggregateMailProvider> logger)
+    : IMailProvider
 {
-    private readonly IOptionsSnapshot<MailConfiguration> _options;
-    private readonly IMailProviderFactory _factory;
-    private readonly ILogger<AggregateMailProvider> _logger;
-
-    public const string FallBackFailedMessage = "No registered mail provider was able to send the message";
-
-    public AggregateMailProvider(
-        IOptionsSnapshot<MailConfiguration> options,
-        IMailProviderFactory factory,
-        ILogger<AggregateMailProvider> logger)
-    {
-        _options = options;
-        _factory = factory;
-        _logger = logger;
-    }
-
     public async Task SendAsync(MailMessage message)
     {
+        message.From ??= options.Value.From;
+        message.FromDisplayName ??= options.Value.FromName;
 
-        if (message.From == null)
-        {
-            message.From = _options.Value.From;
-        }
-        if (message.FromDisplayName == null)
-        {
-            message.FromDisplayName = _options.Value.FromName;
-        }
-        foreach (var providerConfiguration in _options.Value.Providers)
+        foreach (var providerConfiguration in options.Value.Providers)
         {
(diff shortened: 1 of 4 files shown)


jrmccannon (commented):
I'd say this looks fine to me. Might be better if @jonashendrickx gave it a look as well.

src/Common/Services/Mail/AggregateMailProvider.cs
```diff
@@ -5,53 +5,32 @@ namespace Passwordless.Common.Services.Mail;
...
                 return;
             }
             catch (Exception e)
             {
-                _logger.LogError(e, "Failed to send message using provider '{Provider}'", providerConfiguration.Name);
+                logger.LogError(e, "Failed to send message using provider '{Provider}'", providerConfiguration.Name);
             }
         }
-
-        _logger.LogCritical(FallBackFailedMessage);
```
jonashendrickx, line 54: If none of the registered e-mail providers can send, we need to know all of them fail. Because at that point the customers could be critically impacted being unable to log in.

src/Common/Services/Mail/AggregateMailProvider.cs
```diff
@@ -5,53 +5,32 @@ namespace Passwordless.Common.Services.Mail;
...
+                var provider = factory.Create(providerConfiguration.Name, providerConfiguration);
 
                 await provider.SendAsync(message);
-                _logger.LogInformation("Sent message using provider '{Provider}'", providerConfiguration.Name);
+                logger.LogInformation("Sent message using provider '{Provider}'", providerConfiguration.Name);
                 return;
             }
             catch (Exception e)
             {
-                _logger.LogError(e, "Failed to send message using provider '{Provider}'", providerConfiguration.Name);
```
jonashendrickx, line 50: This one was just to let us know when we fail to send using a specific provider, and log accordingly if it is not being logged by the provider itself specifically. It is to help us know when there is a specific issue with either a template or credentials of a specific provider.

src/Common/Services/Mail/MailBootstrap.cs
```diff
@@ -17,26 +17,12 @@ public static void AddMail(this WebApplicationBuilder builder)
             {
                 var section = builder.Configuration.GetSection("Mail:Providers");
 
-                if (!section.GetChildren().Any())
-                {
-                    o.Providers = new List<BaseMailProviderOptions>
-                    {
-                        new FileMailProviderOptions()
```
jonashendrickx, line 24: Removing it might not be a good idea, we could define a new mail provider to log to the console using a `warning` or `info` log level for testing. Or output to Memorycache or some other in-memory option.

For self-hosting keeping the file is better by default.

jonashendrickx: I think all you need to do to fix the API testing image is the following:

https://github.com/bitwarden/passwordless-server/pull/716

jonashendrickx: It is ready, the tests are passing

jonashendrickx:
This would swallow any error messages

jonashendrickx (changes requested):
Would swallow error messages, if magic links fail we would want to know about it.

